PR #376, opened 2015-01-11
maps were crashing when topics were removed

Connoropolous:
(if those topics included the 'root node'). 
this was caused by a setTimeout interfering with synchronicity of events.

the change is pushed to metamaps.herokuapp.com so you can test there. Please make any comments here, not in the asana.

To help with testing, open up developer tools, go to 'console' and run the following 
`Metamaps.Visualize.mGraph.graph.getNode(Metamaps.Visualize.mGraph.root).name`

It will display the name of the node currently set as `root`. Which you can then experiment with removing/ not removing.

closes https://app.asana.com/0/15303030668422/22902140781663

### DIFF
--- a/app/assets/javascripts/src/Metamaps.js
+++ b/app/assets/javascripts/src/Metamaps.js
@@ -2812,10 +2812,6 @@ Metamaps.Control = {
     hideNode: function (nodeid) {
         var node = Metamaps.Visualize.mGraph.graph.getNode(nodeid);
         var graph = Metamaps.Visualize.mGraph;
-        if (nodeid == Metamaps.Visualize.mGraph.root) { // && Metamaps.Visualize.type === "RGraph"
-            var newroot = _.find(graph.graph.nodes, function(n){ return n.id !== nodeid; });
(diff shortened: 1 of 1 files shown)


G1TB3N:
Heroku was still crashing like a hollywood getaway car for me today...anyone else?

Connoropolous:
Ok, yes, but WHEN Sir Ben?

G1TB3N:
Removing a topic from map - presumably not the root (i didn't run the console check yet). This is on Firefox, freshly logged in, have not done a hard cache clear at this time, will  try that next if I'm the only one seeing this issue persist.

G1TB3N:
Closed? As in fixed? Still happening on the test site for me.
A key point to note is that if a map crashes after removing topic, and you don't refresh but instead return to home page and go to another map *THAT* map will open without labels on the topics.

Connoropolous:
It's not going to get merged into either live site until we totally iron it
out, it just made sense to me to get it into the master branch as part of
the work that I'm doing. I'll try and hunt down these ongoing circumstances
where it's still happening.

PR #378, opened 2015-01-29
updated embedly library, and data-card-chrome was now causing a breaking change

Connoropolous:
this closes the recent issue where embedly links were not showing anything besides the image, and if no image, nothing at all.


poietic:
+1!

PR #380, opened 2015-01-29
when a topic was deleted, renderEdgeArrows was trying to call on a topic...

Connoropolous:
... that was no longer existent to draw synapse direction, breaking the graph render


Bortseb:
+R

PR #397, opened 2015-02-19
allow user to choose between map screenshot being full map, or specific view

### LINKED ISSUE #393: Export map images with highlighted synapses
there's already been 3 people who have asked for this, two of them from within the uservoice ticket queue. without it, the export feature's really not that useful - they're defaulting back to taking screenshots.

Connoropolous:
- [x] wire up code
- [ ] add distinct images for capturing 'current view' vs. 'full map'

closes #393

### DIFF
--- a/app/assets/javascripts/src/Metamaps.JIT.js
+++ b/app/assets/javascripts/src/Metamaps.JIT.js
@@ -26,7 +26,12 @@ Metamaps.JIT = {
         };
         $(".zoomExtents").click(zoomExtents);
 
-        $(".takeScreenshot").click(Metamaps.Map.exportImage);
+        $(".takeScreenshotCurrent").click(function(){
+            Metamaps.Map.exportImage('current');
+        });
+        $(".takeScreenshotFull").click(function(){
+            Metamaps.Map.exportImage('full');
+        });
 
         self.topicDescImage = new Image();
         self.topicDescImage.src = '/assets/topic_description_signifier.png';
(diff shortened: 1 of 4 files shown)


Connoropolous:
@poietic it's pushed up to heroku. There's something wrong with the S3 config with heroku to have that part work, and make the actual images save, but if you hit download, you'll see exactly what would save if it was working. I've got the part working locally that will actually make it work on live :) 

just go hover over the 'capture screenshot' icon

poietic:
@Connoropolous can you describe to me what you think is the rationale behind enabling both different possibilities? my consideration is that we may be adding an extra button where there doesn't need to be.

perhaps we just stick with the "capture current view" button, and let go of the "capture whole map". "Capture whole map" only makes sense if the image that it captures is high-resolution enough to be read, but at this stage it isn't - so its not a whole lot different than just zooming out and capturing that view.

that being said we're mixing things up a bit because clicking the capture button sets the map thumbnail...so if we simplified this, it  would mean that whatever view people captured would set the map thumbnail...which i'm not sure is 100% desireable...

have to resposition cafes now, will come back to this in a bit.

devvmh:
+1 to ishan's comment above - I think it's a more elegant solution to have one button; users can zoom out if they want the whole map view. We're probably better off having a "zoom to see full map" button somewhere else in a utility dropdown

Looking at the code it's not that heavy... but I suppose every extra complexity is more complexity. So yeah, I vote this is unneeded

Connoropolous:
Closing for the reasons mentioned above. we'll open another one for this, implemented in the way described above

PR #400, opened 2015-03-01
Add xls and csv export view to map

devvmh:
So that if you visit e.g.

http://www.metamaps.cc/map/9999.csv or
http://www.metamaps.cc/map/9999.xls

you'll get a CSV or XLS export.


devvmh:
https://github.com/metamaps/metamaps_gen002/pull/442

PR #402, opened 2015-03-02
Import from CSV feature

devvmh:
This feature isn't done so I'm not sure if it makes sense for it to be a pull request - maybe it shuold be an issue.

I've implemented a proof of concept, basically, for importing data from Excel. You can create a two-column table in Excel of name/metacode, and then copy-paste it into a textarea on the site and run a function, which generates all of the topics on the map instantly.

There are a few things to do to it before it will be workable.

1) Think more about the format of the input from Excel
2) Think about how to incorporate this magical textarea to accept the input, plus making a button to activate the feature, plus thinking about where to put that in the interface and how to explain/document it
3) Come up with a smarter algorithm for placing the topics on the map (not necessarily that much smarter, but I was thinking of offsetting rows so they sort of came in a checkerboard pattern, and then also it should respond to browser window size better)
4) Other things I haven't thought of

### DIFF
--- a/app/assets/javascripts/src/Metamaps.js
+++ b/app/assets/javascripts/src/Metamaps.js
@@ -4053,6 +4053,73 @@ Metamaps.Topic = {
             }
         }
     },
+    createCSVInputTextarea: function() {
+      $('body').append('<textarea id="parseCSV" style="position: absolute; top: 50%; left: 50%;"></textarea>');
+      $('#parseCSV').keyup(function(e) {
+        if (e.keyCode == 13) { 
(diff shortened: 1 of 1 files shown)


devvmh:
Another thing we migth do if accepting this commit is to use a call to createTopicLocallyFromArguments to simplify the createTopicLocally function, to have less duplicated code

Bortseb:
For reference:
http://stackoverflow.com/questions/51783/how-to-serialize-a-graph-structure

devvmh:
https://github.com/metamaps/metamaps_gen002/pull/443

PR #406, opened 2015-03-03
extract email settings into ENV :scissors:

s01us:
here you go! 
:gift:

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -25,10 +25,11 @@ Metamaps::Application.configure do
     
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.smtp_settings = {
-    address:              'mail.metamaps.cc',
-    port:                 587,
-    user_name:            'team@metamaps.cc',
-    password:             'RcxX+s:fht49UX',
+    address:       ENV['SMTP_SERVER'],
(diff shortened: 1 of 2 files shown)


config/environments/development.rb
```diff
@@ -25,10 +25,11 @@
...
   config.action_mailer.smtp_settings = {
-    address:              'mail.metamaps.cc',
-    port:                 587,
-    user_name:            'team@metamaps.cc',
-    password:             'RcxX+s:fht49UX',
+    address:       ENV['SMTP_SERVER'],
+    port:          ENV['SMTP_PORT'],
+    user_name:     ENV['SMTP_USERNAME'],
+    password:      ENV['SMTP_PASSWORD'],
+    #domain:        ENV['SMTP_DOMAIN']
```
Connoropolous: Looks good, but what's this line for?

s01us: My understanding is its a setting to declare the where the mail is coming
from, to increase deliverability / minimize bounce.
May only be relevant if you have a separate mail server

http://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration

Let me know if you want it out.

Connoropolous: Ok, that makes sense. No it can stay. I'll merge this in!

PR #408, opened 2015-03-08
add cancancan

mixmix:
cancan is no longer supported by ryan bates
the community has picked up the project and is maintaining the project under the gem [cancancan](https://github.com/CanCanCommunity/cancancan)

the syntax is all the same (still uses the same classes and methods), you don't need to change anything


devvmh:
Hey @mixmix thanks for this. I've included this change in my update to rails 4 branch, which will be merged into develop and then eventually into master :)

PR #415, opened 2015-09-30
only autocomplete topic/synapse if the user types more than 2 characters

devvmh:
This branch only starts autocompleting if there's more than one character typed in the new topic/synapse box. I hope this PR can be used to discuss if

a) should we have this feature?
b) should it be configurable per instance? I'm not currently sure how to set this value based on an environment variable, but it should be doable
c) what is the ideal number for metamaps.cc?


Connoropolous:
This makes a lot of sense to me. @poietic what has your experience been?

devvmh:
I feel cool pulling this - it's easy to change in the future and 3 people are on board

PR #420, opened 2015-10-28
Realtime video and chat

### LINKED ISSUE #444: HTTPS problems
Modern browsers (particularly Chrome) freak out if you serve HTTP content on a page that's supposed to be secured. This means that currently if you access metamaps via https://metamaps.cc, it won't work. There are several issues to solve here.

Further, for realtime video to work in Chrome, metamaps MUST be served over HTTPS. So our hand is being forced a bit.

Here's what we need to do:

 - [x] Ensure all rails-internal assets are served over either https or "//", which will infer the protocol. Since we're moving long term to https-only (which is a good decision I think, even though it's a bit harder to support serverside), it's probably best to use https wherever possible (since we can serve https assets safely inside the http page. I can take this on, progress is in the https://github.com/metamaps/metamaps_gen002/tree/feature/https-preparation branch
 - [x] http://support.metamaps.cc/logout.json gives an error, but it's required in our page. @Connoropolous and/or @poietic do you have any insights? The answer may have to be changing the url to metamaps.uservoice.com, per [this page](http://feedback.uservoice.com/knowledgebase/articles/16819-site-settings-for-uservoice-product-management#forcessl).
 - [x] http://metamaps.cc:5001 gives an error. I suggest creating a CNAME called realtime.metamaps.cc, so we can change the javascript code to https://realtime.metamaps.cc (which will proxy to http://127.0.0.1:5001, so it'll end up being the exact same realtime server except using https).
 - [x] Ensure all user profile images are served over HTTPS, or they won't be visible. This is harder. This entails either enforcing https on the specified images, or using file uploads instead. I'm not 100% sure but it's possible these are already served via AWS, which has HTTPS so we're good. I'll check this out and make sure it's AWS or tag people to help me decide how to handle this. Used the s3_protocol configuration option on paperclip to fix map screenshots and presumably user profile images too.
 - [x] Ensure all custom metacode icons are served over HTTPS. Currently there are a lot hosted at dropbox (on staging); I'm not sure if these are HTTPS or not. Anyways, these ones we definitely need to change our strategy. I'm kind of annoyed by the UX issue of forcing users to use HTTPS icons only. Probably the best long term solution is to allow arbitrary URLs, but then grab them and upload them to AWS if they're HTTP. @Connoropolous @poietic @benb I welcome input on this.
 - [x] ~~Consider linking to ssl-enabled versions of blog.metamaps.cc and support.metamaps.cc. This isn't 100% required but I think it's a good idea? @poietic I'll need your help with the blog.metamaps.cc part of this, since I don't know how that's hosted. If you'd like, we could move this to the linode server on metamaps.cc, but if we can keep it on web hosting hub it means less work for me in the long run.~~
 - [x] Push the changes to the production server, but don't enforce SSL yet. Then we can do testing on https://metamaps.cc while still serving the content at http://metamaps.cc Once we're confident, we can start enforcing http. [EDIT: well, this isn't done, but it can be part of the rails4 push]

This ticket is a bit timely if we want real time video to work well when it launches.

Connoropolous:
- [x] localVideo
- [x] others videos
  - [x] display user color as border around video element
- [x] show avatar instead of video feed for others, when you turn your own off
- [x] z-indexes
- [x] remove the old Junto panel
- [x] the signal server
- [x] the chat panel
  - [x] set up the participants box
- [x] merge #448 and fix #444 so https is ready
- [x] fix double message issue in chat


Connoropolous:
Almost there. Today @devvmh reported seeing two of each message that I was sending, while we were on a map together
- [ ] fix it

Connoropolous:
@devvmh I believe this branch is ready to go in =D

Connoropolous:
@devvmh my only ? is that I can see a bit of code relevant to CSV and XLS showing up in my branch, that must not be in develop... I'm assuming it was in develop, I merged develop into this branch, and then you took it out? Is that the case? If so, should we remove it from this branch again?

devvmh:
Uhhh.... yeah that's weird, but no worries. If you want just merge it, it shouldn't make any UI difference. I'll rebase my branch after the push.

Connoropolous:
Ok. I'm going to merge it folks I'm going to merge it :shipit:

PR #440, opened 2015-11-30
fix filter bug

devvmh:
It looks like the if statements in this section of the file are the issue. Since "" evaluates to false, they don't get added to the filters list. This means if you have a few "" synapses in your map, and they're already correctly in the list, updateFilters will REMOVE them from the list when you add a new synapse.

This code fixes that problem. I believe I already fixed the forking problem elsewhere in the rails 4 code. We can test that assertion during our overall testing

@Connoropolous could you check this and merge it please?


Connoropolous:
Yes, I'll review and merge

Connoropolous:
+R

PR #456, opened 2016-01-06
rspec tests

devvmh:
This branch enables rspec testing. It's far from done, but I want to merge it so that we can start using rake db:seed instead of rake db:fixtures:load.

Currently this branch tests all of the models fairly thoroughly. I hope to also test the controllers, so there are 77 "pending" tests when you run rspec.

Once this is done (and we're done realtime video, etc) I hope to integrate javascript tests into this as well!

As I built the model tests, I added some validations and factored some code out of model files.

The code diff of this PR is most interesting if read from the bottom up.


harlantwood:
Great work!

config/application.rb
```diff
@@ -45,6 +45,9 @@ class Application < Rails::Application
     # Version of your assets, change this if you want to expire all your assets
     config.assets.version = '2.0'
 
+    config.generators do |g|
+      g.test_framework :rspec
```
Connoropolous: does this run the tests?

app/models/user.rb
```diff
@@ -26,7 +26,7 @@ class User < ActiveRecord::Base
   validates_uniqueness_of :name # done by devise
   validates_uniqueness_of :email # done by devise
 
-  validates :joinedwithcode, :presence => true, :inclusion => { :in => $codes, :message => "%{value} is not valid" }, :on => :create
```
Connoropolous: So maybe we can get rid of the $codes global variable then? I have to find it again...

Connoropolous:
I mostly read this over to check the refactoring and the switch from fixtures to seeds, as the amount of code you've written for tests is a bit more than I have bandwidth for at the moment, and it also wouldn't break anything if there was something wrong. 

It would be great to be able to have javascript tests as well, however, I wouldn't go forward with that at all any time soon because the javascript code is, I believe, still too unorthodox to merit having tests written for it, when it may be that we switch to a more stable and structured front end framework like angular or who knows what in the mid-term future...

+R great work Devin!

app/models/user.rb
```diff
@@ -26,7 +26,7 @@ class User < ActiveRecord::Base
...
+  validates :joinedwithcode, :presence => true, :inclusion => { :in => User.all.pluck(:code), :message => "%{value} is not valid" }, :on => :create
```
Connoropolous: aha... no, I don't think we should switch this... there is some kind of situation that I would run into, I forget precisely what it was now, where ActiveRecord hasn't yet been initiated so that this call to the `User` model will work, which is why I had to go directly to a db connection in https://github.com/metamaps/metamaps_gen002/blob/6767771d7d0c49dc6bdacef1a8f126446db7d397/config/initializers/access_codes.rb

devvmh: OK solid I'll switch it back to $codes. I don't remember why I switched it, probably just because I didn't understand it.

devvmh:
I agree - the seeds bit is really the only important part. Tests are sort of orthogonal right now. Trust me, though, they're really great :D

Once I've fixed the $codes thing, I'll merge this.

devvmh:
Thanks @harlantwood :+1:

config/application.rb
devvmh: rspec runs the tests, I probably just got that line from documentation somewhere. https://github.com/metamaps/metamaps_gen002/blob/develop/doc/RspecTesting.md explains how to run the tests - it just means executing `rspec` in the top level directory.

devvmh:
@Connoropolous re: the javascript testing... what really motivates me is this concept I learned about "legacy code"; some bloggers online define it as 'code without tests'. It's a big job to dive into the thousands of lines of Javascript that makes up metamaps, but with tests, in theory, the refactoring should be even easier. There's lots to think about there, but it's a long way off anyways.

OK, I'm merging this into develop

PR #460, opened 2016-02-07
xls import: bare minimum functionality - use by Ctrl+V onto the map canvas

devvmh:
I'm not sure if this should be merged yet, but here it is.

Copy paste text from Excel onto the map canvas. A prompt should appear asking if you want to create "n" topics, where n is the number of rows (less the header row).

It should be able to treat the topics like a csv, and as long as you have metacode and name it should be able to create something workable.

I used the format that is exported by feature/xls.export, but this should hopefully be more flexible than that.

TODO:

- [ ] finish synapse import
  - [ ] mappings aren't persisted
- [ ] finish topic import
  - [ ] xloc and yloc aren't taken up

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -22,6 +22,7 @@
 //= require ./src/Metamaps.Views
 //= require ./src/JIT
 //= require ./src/Metamaps
+//= require ./src/Metamaps.Import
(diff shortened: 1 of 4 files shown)


devvmh:
Please see discussion in #475 for the next steps for import

PR #473, opened 2016-02-19
fix get_invite_link function. fixes #471

### LINKED ISSUE #471: fix invite link
Invite links used to mandate coming from http://metamaps.cc. I whipped up some code to make it automatically infer the url from the request, and fall back to //metamaps.cc if it couldn't do it.

Two problems:

1. //metamaps.cc looks dumb, and doesn't work as a shared invite link
2. the code isn't working, so metamaps.cc is currently serving invite links that look dumb

Fixes, in order of desirability:
- [ ] Make the code smarter so it can infer the hostname sucessfully.
- [ ] introduce an env variable that specifies the hostname.
- [ ] Update the code so it specifies metamaps.cc/whatever instance manually again.

devvmh:
that rascally request.base_url function couldn't evade me forever

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -64,9 +64,6 @@ private
   end
 
   def get_invite_link
-    unsafe_uri = request.env["REQUEST_URI"] || 'https://metamaps.cc'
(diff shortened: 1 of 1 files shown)


Connoropolous:
There we go! Easy. +R

PR #474, opened 2016-02-21
don't need xloc or yloc for Synapse mappings

devvmh:
Fix baaaad problem on develop branch with new Synapse mapping validations, that prevented new Synapses from staying associated with the map they were on.


Connoropolous:
Good catch. What is Proc.new ?

devvmh:
Proc.new is basically a shorthand function. There's a difference between a Proc and a lambda and a block, but I'd have to look it up to tell you the difference.

It's essentially copy-pasta from http://guides.rubyonrails.org/active_record_validations.html#using-a-proc-with-if-and-unless

PR #477, opened 2016-02-26
metacode icons on aws (fixes #445)

### LINKED ISSUE #445: Metacode icons not at an HTTPS url won't work if Metamaps is served over https
### Game plan (made easier since metacode editing is only for admins):

- [x] Create a database migration that changes icon into an aws field
- [x] Make sure that metamaps.cc icons will be safely migration from a text field like "/assets/blueprints/bp_wildcard96px.png" to being a proper paperclip field
- [x] Metacode paperclip file upload UI instead of a text field
- [ ] ~~Migrate metamaps.cc~~
- [ ] ~~Ensure local dev instances and staging work OK~~
- [ ] ~~Ensure other instances are/will be migrated OK~~

(I'm happy closing this issue before we migrate every instance)

### OLD EXPLANATION:

Once we move to using https only, we'll have a UX issue.

If I want to create a new metacode set using images hosted at http://insecure.devinhoward.ca, which doesn't have SSL, I'll have a problem. I'll create the metacodes, and for the icon field I'll fill in a url like "http://insecure.devinhoward.ca/cool-metacode.png". But then Chrome and Opera and others will block that content, or worse, mark the entire page as insecure because there's an insecure image in the page for every page containing that metacode.

So possible solutions are:

1. Only allow the icon field to take urls starting with https
2. Change the field to an AWS file upload like user profile images [EDIT: going with this one]
3. Keep the field as a text field, but if it starts with http, use a script to download that image and then store it on AWS and serve it via HTTPS.

Here are the problems I've noticed for each approach:

1. This kind of sucks as as solution, because I might put https://insecure.devinhoward.ca/cool-metacode.png in the field, and then it won't be secure, so it'll create security errors.
2. Perhaps we didn't do this in the past because of copyright issues? I'm not sure. This seems like the best solution to me.
3. This is ambitious, lots of work, and still has problems. If a website starts out secure, but the certificate expires, we're screwed. I really wouldn't want to be tracking down these insecure images and fixing them later...

After reflecting on these three options, I've convinced myself the only way is to do it all via AWS and serve it over HTTPS, because that's really the only reliable way I see to ensure all metacode icons are secure forever. Do others have thoughts?

devvmh:
This branch changes old metacodes (and the metacodes in db/seeds.rb) to live in a field called "manual icon" that enforces https on its urls. New metacodes can be uploaded through the web interface and will be stored on AWS using Paperclip. Metacode icons will be automatically deleted if the metacode is deleted or the icon replaced.

Manual icon can only be filled through the rails console, db seeds, or directly manipulating the database.

Metacode icons can still be accessed via `@metacode.icon` as before. It may be better to access them by `@metacode.icon(:ninetysix)`, I'm not sure.

A database migration included in this branch should migrate all instances seamlessly, UNLESS they have metacodes with non-https urls. The migration will go OK, but we need to get into the interface right away to update those urls to HTTPS. One option is downloading all of the icons, and then re-uploading them through the edit metacode form.

@Connoropolous, I'll wait for a final go-ahead from you before I merge this.

Once this is deployed it will close #445.  The only remaining HTTPS issue will be with embedded links :(:(:(:(:(


Connoropolous:
nice! Love the css refactoring. Although the css is another example of a front end nightmare, as it doesn't  use any of the scalable and modular css approaches http://www.sitepoint.com/css-architectures-scalable-and-modular-approaches/

Ah well, it will all get overhauled in due time, in a front end rewrite. For now, we keep getting the back end sorted out.

PR #480, opened 2016-02-29
fix #479

### LINKED ISSUE #479: js bug in develop
https://github.com/metamaps/metamaps_gen002/blob/develop/app/assets/javascripts/src/Metamaps.js.erb#L20

this is overriding the current value of the Metamaps global, which has already started to be defined and shouldn't be reset here. each of these 4 should be set individually

(logging for myself or for @devvmh to do at earliest convenience)

devvmh:


Connoropolous:
you're too fast @devvmh =D 

there's a bit more though, tack another commit on there with a check in this function to make sure that Metamaps.whatever is an object 
https://github.com/metamaps/metamaps_gen002/blob/develop/app/assets/javascripts/src/Metamaps.GlobalUI.js.erb#L50
`typeof (Metamaps[prop]) == 'object'`

Connoropolous:
ah dangit. that doesn't work. 

`typeof null` is equal to `object`

devvmh:
Haha dang typeof [] also is "object"

Connoropolous:
`Metamaps[prop] != null` works

Connoropolous:
Best thing would be for you to try and run the app with that. Seems to me a thing or two might still be unhappy. Including the comments using `#` which I don't think are allowed.

devvmh:
Whoops, changed the comments. I tried running it quickly but I'll maybe test this later

PR #484, opened 2016-03-13
more token entropy

harlantwood:
Because of limited character set and some characters fixed by the UUID spec, UUID gives us way less entropy than base 64.  We like entropy :+1: 

We create a loop in case a candidate token is shorter than 32 chars.  None were when I tested a random sample of 1 million, but theoretically a token could be of length zero, which would be bad ; ) 

I split it into two methods for ease of testing.

### DIFF
--- a/app/models/token.rb
+++ b/app/models/token.rb
@@ -1,11 +1,22 @@
 class Token < ActiveRecord::Base
   belongs_to :user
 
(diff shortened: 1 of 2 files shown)


Connoropolous:
Great! Thanks Harlan. Will take another look and test tomorrow and see
about merging it

PR #488, opened 2016-03-14
no cancan

devvmh:
remove cancancan/cancancan gems, since we have pundit now


Connoropolous:
Actually @devvmh the gem Snorlax relies on our having included cancan, even though we don't use it, so we should close this

devvmh:
Good point, but I just checked and snorlax relies on cancancan. We don't
need the old/outdated cancan gem in our Gemfile; snorlax will include what
it needs just fine

Although it'd be great if Snorlax didn't rely on cancancan....

Connoropolous:
alrighty, merge it then :)

PR #490, opened 2016-03-15
call for developers in Inspect Element window

devvmh:
i went with the "learn Ruby/Javascript" angle. We could reword this.

@poietic @Connoropolous @G1TB3N thoughts?


Connoropolous:
@devvmh I love it! We could write the same string to the JS console as well :p

Connoropolous:
If it's running in a Production environment

PR #496, opened 2016-03-23
update version, date, and active peers

Connoropolous:
@devvmh I set the version as 2.9 because I am thinking that once we merge Junto, that's what it could be


devvmh:
Currently on metamaps.cc the version is 2.7. I incremented it right after the last push. I'll switch it back to 2.8.

Connoropolous:
ok, coo

PR #499, opened 2016-03-25
use new pundit-enabled snorlax

devvmh:
Well, one pull request later, snorlax 0.1.5 solves our problem in a much more elegant way.

@Connoropolous plz still test :+1: :+1: :+1:


Connoropolous:
Great great great. This combined with the other PR is perfect

Issue #475, opened 2016-02-21
better import/export


devvmh:
Once we merge feature/xls.import and feature/xls.export, here's what I'd like to enable on top of the current functionality:

**Released in v2.8**

- [x] enable xls/csv export
- [x] enable copy-paste import
- [x] firm up all map API endpoints
- [x] consider permissions during import/export
- [x] make x and y optional during import
- [x] metacodes SHOW api endpoint
- [x] improve the serialization format for csv/xls
- [x] implement json export
- [x] re-implement import with the new format
- [x] add a json parser to the JS import module

**(will be) Released in 2.10**

- [x] drag and drop files triggers import
- [x] add a csv parser to the JS import module
- [x] icon on top right for import with
  - [x] instructions on copy-paste
  - [x] instructions + button to export map
  - [x] file upload field for csv files
- [ ] browser testing

** Decided to drop **

- [x] ~~upgrade tsv parser in the JS import module~~
- [x] ~~file upload parser in the JS import module for xls/csv/ods/xlsx?~~

harlantwood:
Awesome to see the movement on this!  

Related: #431

devvmh:
Oh that's a good point.

I never really understood what JSON-LD meant, but this is very linked. Looking at https://github.com/metamaps/metamaps_gen002/compare/feature/json makes me realize it's very similar but @Connoropolous  had some better ideas going on.

I think one of the biggest issues for me is the serialization format. A recent commit made the xls and csv have the same serialization format, but really, a map is hard to represent. Currently, I've got it in a spreadsheet because I have a certain workflow in mind - transferring between Excel/Libreoffice and a map to do Excel operations. 

But I'm not sure that it's the best representation. It does break the csv spec a bit, because there are two headers, and it makes the import code more complicated.

I know @Bortseb has thoughts about import/export, but we need more than thoughts, we need proposals! There is a comment at the top of https://github.com/metamaps/metamaps_gen002/blob/feature/xls.import/app/assets/javascripts/src/Metamaps.Import.js.erb on the xls.import branch that describes the current input/output format from my two pull requests.

JSON is probably a more efficient/natural way of representing a map, since you can make it a list of topics/synapses.

In working on xls.import, I also discovered that you do need a concept of an "import id". That's the only way to describe synapses. But then you need to map that import id to an actual new id in the database, because you're creating new topics.

[...]

PR #506, opened 2016-03-26
minimal import/export - could go in v2.8 for the keeners

devvmh:
OK, this matches the spec in #475. You can export map #999 from these urls:

    GET /maps/999.csv # redirects
    GET /maps/999.xls # redirects
    GET /maps/999/export.json
    GET /maps/999/export.csv
    GET /maps/999/export.xls

Additionally, `GET /maps/999.json` retains its old behaviour.

metacodes#show is implemented, taking a name or id as the parameter.

The export format is changed to match the spec in #475.

The import now allows tab-separated values or JSON (more formats on the way in another branch for another release), and has been changed to use the new format from #475.

This doesn't change the UI in any way, except for intercepting the results of someone copy-pasting text onto the canvas/body of a map. I don't think anyone will notice unless they're looking for it.

Future versions of this functionality will have a UI to make it simpler for users.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -25,6 +25,7 @@
 //= require ./src/views/room
 //= require ./src/JIT
 //= require ./src/Metamaps
+//= require ./src/Metamaps.Import
 //= require ./src/Metamaps.JIT
 //= require_directory ./shims
 //= require_directory ./require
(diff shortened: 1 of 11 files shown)


devvmh:
One bug I've noticed is that synapses are invisible after import until you reload the page :(

Connoropolous:
@devvmh given that there's a minor bug on the OAuth flow as well, on a functionality that I don't expect anyone to find or use within the first week, I would also advocate this going in. Do eet. For myself, I'm going to be sure to address the OAuth bug within the week

Connoropolous:
+R

PR #535, opened 2016-04-05
add rack-cors; fix #531

### LINKED ISSUE #531: Browsers should be able to access metamaps API from JS running on any domain (CORS)
specifically, I believe we should add these headers for API responses:

```
Access-Control-Allow-Methods: POST,GET,PUT,DELETE,OPTIONS,HEAD
Access-Control-Allow-Origin: *
```
 
@Connoropolous this will remove the need for the `corsHackUrl` in the [Metamaps Nodesphere adaptor](https://github.com/nodesphere/nodesphere/blob/2096736ab46c878a140b645dd3789933881edb0f/src/adaptor/metamaps.coffee#L11) descended from the one we worked on together.

@Connoropolous and @devvmh can you think of other HTTP verbs that are used for the API?

devvmh:


harlantwood:
Awesome!!  Thanks Devin : )

Connoropolous:
Thing is, I think it is a security concern to have this enabled on endpoints that aren't the API, so maybe we can find a way to limit which routes use the middleware

devvmh:
Good point. I really don't understand CORS myself, beyond a sort of fuzzy understand at the edge of my awareness

Based on the docs, this looks like it'll limit CORS stuff to routes beginning with /api, which should be about right.

I've now read about people's asset serving being screwed up by this, so our next deploy should be a careful one.

Connoropolous:
+R

harlantwood:
Yeah, maybe deploy separately from anything else esp migrations so it's easy to roll back if needed.

Connoropolous:
this will close this: https://github.com/metamaps/metamaps_gen002/issues/531

@devvmh will you merge and deploy?

PR #536, opened 2016-04-06
metamaps.debug function, and move VERSION into a rails constant initializer.

devvmh:
from this commit on, we would need to modify config/initializers/version.rb to increment the version number.


Connoropolous:
Nice!

PR #546, opened 2016-04-13
slack exception notifications

devvmh:


Connoropolous:
Cool! So we create a channel in our metamaps team for this, and then create an integration, and then send error logs to that hook?

devvmh:
Exactly. The docs are on github in the gem's repo.

I just figured we're running out of integrations!

Connoropolous:
did you test it?

devvmh:
no, and again, because we're running out of integrations

i also put this in "needs discussion" because i'm not even sure we want it. But it is pretty cool at work having a channel that is a feed of production errors, since we can watch it and go squash bugs as they come up

I'm not super motivated to move fast on this though, I'm using bits of time here and there for JS refactoring right now

devvmh:
We could merge this code and then not even use it, I think.

Connoropolous:
Yes, that's true, we could

Connoropolous:
We can also repurpose other webhooks as well, no need to always create a
new integration

PR #547, opened 2016-04-14
added an event for conversations starting

### LINKED ISSUE #539: Implement 'conversation_starting' slack message
steps: 

- [x]  add `conversation_started_on_map` to here: https://github.com/metamaps/metamaps_gen002/blob/develop/app/models/event.rb#L2
- [x] make a copy of  https://github.com/metamaps/metamaps_gen002/blob/develop/app/models/webhooks/slack/topic_added_to_map.rb and rename it to `conversation_started_on_map.rb`
- [x] edit the class name here https://github.com/metamaps/metamaps_gen002/blob/develop/app/models/webhooks/slack/topic_added_to_map.rb#L1 from `Webhooks::Slack::TopicAddedToMap` to `Webhooks::Slack::ConversationStartedOnMap`
- [x] change the message that gets sent to slack here: https://github.com/metamaps/metamaps_gen002/blob/develop/app/models/webhooks/slack/topic_added_to_map.rb#L4
- [x] add a line in `routes.rb` right after https://github.com/metamaps/metamaps_gen002/blob/develop/config/routes.rb#L46 that looks like `get 'maps/:id/events/conversation_started', to: 'maps#conversation', as: :conversation
- [x] in `maps_controller.rb` add a new `action` definition called `conversation`, modelled after https://github.com/metamaps/metamaps_gen002/blob/develop/app/controllers/maps_controller.rb#L106, but make it include a line like this: https://github.com/metamaps/metamaps_gen002/blob/develop/app/controllers/mappings_controller.rb#L25
- [x] create a new file in the `app/models/events` directory called `conversation_started_on_map.rb`, and copy the code from https://github.com/metamaps/metamaps_gen002/blob/develop/app/models/events/new_mapping.rb into it. 
- [x] modify line 5 of that file to: `create!(kind: "conversation_started_on_map",`
- [x] from within the `callAccepted` function, make an jquery ajax GET to the route `/maps/MAPID/events/conversation` https://github.com/metamaps/metamaps_gen002/blob/develop/app/assets/javascripts/src/Metamaps.js.erb#L2314, like `$.get( "/maps/3/events/conversation")`. This will cause the event to fire, and slack to be notified, if there is a webhook attached to that map

Connoropolous:
closes #539

### DIFF
--- a/app/assets/javascripts/src/Metamaps.js.erb
+++ b/app/assets/javascripts/src/Metamaps.js.erb
@@ -2270,7 +2270,7 @@ Metamaps.Realtime = {
         invited: Metamaps.Active.Mapper.id,
         inviter: userid
       });
-      $.get('/maps/' + Metamaps.Active.Map.id + '/events/conversation');
+      $.post('/maps/' + Metamaps.Active.Map.id + '/events/conversation');
       self.joinCall();
       Metamaps.GlobalUI.clearNotify();
     },
--- a/app/assets/javascripts/src/Metamaps.js.erb
+++ b/app/assets/javascripts/src/Metamaps.js.erb
@@ -2270,6 +2270,7 @@ Metamaps.Realtime = {
         invited: Metamaps.Active.Mapper.id,
         inviter: userid
       });
+      $.get('/maps/' + Metamaps.Active.Map.id + '/events/conversation');
       self.joinCall();
       Metamaps.GlobalUI.clearNotify();
     },
--- a/app/controllers/maps_controller.rb
+++ b/app/controllers/maps_controller.rb
@@ -1,7 +1,7 @@
 class MapsController < ApplicationController
 
-    before_action :require_user, only: [:create, :update, :screenshot, :destroy]
-    after_action :verify_authorized, except: [:activemaps, :featuredmaps, :mymaps, :usermaps]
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/src/Metamaps.js.erb
```diff
@@ -2270,6 +2270,7 @@ Metamaps.Realtime = {
         invited: Metamaps.Active.Mapper.id,
         inviter: userid
       });
+      $.get('/maps/' + Metamaps.Active.Map.id + '/events/conversation');
```
devvmh: I think this should be $.post, to be more restful

app/controllers/maps_controller.rb
```diff
@@ -102,6 +102,24 @@ def export
       end
     end
 
+    # GET maps/:id/events/:event
```
devvmh: so change this

config/routes.rb
```diff
@@ -37,14 +37,15 @@
   
   resources :maps, except: [:index, :new, :edit]
   get 'maps/:id/export', to: 'maps#export'
+  get 'maps/:id/events/:event', to: 'maps#events'
```
devvmh: and this

PR #549, opened 2016-04-20
fix map forking (fix #540)

### LINKED ISSUE #540: Fork map / save to new map is broken currently
Tried this on two maps, via two different users, on two browsers, all getting the "working..." notification briefly but then no result. Connor noted a 'server error' mssg, and will check the log

devvmh:


Connoropolous:
oooooooo

+R 

good investigating monsieur!

PR #552, opened 2016-04-25
user presence event

Connoropolous:


devvmh:
is this in progress for reals?

app/assets/javascripts/src/Metamaps.Visualize.js
```diff
@@ -143,6 +143,9 @@ Metamaps.Visualize = {
       self.mGraph.graph.empty()
     }
 
+
+    if (self.type == 'ForceDirected' && Metamaps.Active.Mapper) $.post('/maps/' + Metamaps.Active.Map.id + '/events/user_presence')
```
devvmh: we should have a better way to say "Map View" or "Topic View"

Maybe something in Metamaps.Settings or Metamaps.Router?

Are there lots of spots in the code that check self.type == ForceDirected?

Connoropolous:
The in progress label just adds itself

app/assets/javascripts/src/Metamaps.Visualize.js
Connoropolous: No, it doesn't make that check too many places. The better check might actually be for Metamaps.Active.Map which *should* never be set if there is no active map. And then same with Metamaps.Active.Topic

devvmh:
haha yeah exactly, so is this actually in progress or is it ready?

PR #556, opened 2016-06-14
brakeman

devvmh:
Add the brakeman gem, fix warnings it generated, and include it in the travis config so it will scan for possible security issues for us when we commit new code.


harlantwood:
Nice!

PR #563, opened 2016-06-30
Update code style automatically using rubocop gem

devvmh:
Rubocop gives lots of warnings, and we don't have to follow all of them.

Someday we could configure rubocop to enforce consistent style, and then fix all the remaining warnings


Connoropolous:
@devvmh are there are any risks associated with this one? did you try running the servers and everything with all these changes?

devvmh:
The automatic updates have no risk. I should recheck that last commit though.

The bigger question is if we want to use rubocop for code style at all. But this pr will help with just general consistency in the short term  (e.g. indentation)

devvmh:
Yea ive decided there's no risk except with the changes to app/controllers/api/tokens_controller.rb - do you think its ok to change the pundit error? I never set myself up to test the api

PR #564, opened 2016-07-01
[WIP] rails 5 and sprockets 4

devvmh:
Update to rails 5 and sprockets 4. Major changes are:

- Gemfile is updated, and ruby-version changes to 2.3.0
- changes for sprockets 4: add a manifest file and fix a module.exports javascript error
- copy the config directory from a blank rails 5 app and merge the changes into our directory
- fix one ActiveRecord error in a controller (I think due to rails 5 changes)
- Remove `embed :ids, include: true` from serializers, because it's not supported in the new version of active model serializers

### Motivation

1. ActionCable, for better web sockets integration
1. ...maybe es6 with sprockets, but it doesn't actually seem too tempting anymore

#### TODO:

- [ ] Fix the serializers so the API format remains the same (it's very different right now)
- [ ] QA testing
- [ ] Make Travis work again


harlantwood:
Nice. Ruby 2.3 has some sweet features :)

PR #565, opened 2016-07-02
fix 2 bugs on develop

devvmh:
mapping policy errors out if mappable is nil

javascript error introduced by me foolishly using module.exports


Connoropolous:
+R !

PR #568, opened 2016-07-24
use Wildcard if no metacode

devvmh:


Connoropolous:
+R makes sense to me and looks good in terms of code :)

PR #573, opened 2016-07-31
fix synapse autocomplete duplicates

devvmh:


devvmh:
ugh should be develop

PR #574, opened 2016-07-31
fix synapse autocomplete duplicates

devvmh:
there were 2 issues. one is synapses with leading/trailing whitespace. I've run a script on production to strip existing leading/trailing whitespace from synapse descs.

The second issue was that this code wasn't actually modifying the @synapses array, and collectedDesc wasn't doing anything. So the new line in main_controller.rb will be more effective than the old code.


Connoropolous:
nice +R

PR #576, opened 2016-07-31
move frontend code into a folder, and setup react build tooling with es6 transforms

devvmh:
Take a look, let me know what you think.

### DIFF
--- a/frontend/.babelrc
+++ b/frontend/.babelrc
@@ -1,0 +1,9 @@
+{
+  "presets": [
+    "react",
+    "es2015"
+  ],
+  "plugins": [
+    "transform-class-properties"
+  ]
+}
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,6 +3,8 @@
   "version": "1.0.0",
   "description": "Metamaps frontend - currently just tests",
   "scripts": {
+    "build": "webpack",
+    "build:watch": "webpack --watch",
     "test": "mocha test || echo 'Run `npm install` to setup testing'"
   },
   "repository": {
@@ -16,7 +18,18 @@
   },
   "homepage": "https://github.com/metamaps/metamaps#readme",
   "devDependencies": {
+    "babel-cli": "^6.11.4",
+    "babel-loader": "^6.2.4",
+    "babel-plugin-transform-class-properties": "^6.11.5",
+    "babel-preset-es2015": "^6.9.0",
+    "babel-preset-react": "^6.11.1",
     "chai": "^3.5.0",
-    "mocha": "^2.4.5"
+    "mocha": "^2.4.5",
+    "path": "^0.12.7",
+    "webpack": "^1.13.1"
+  },
+  "dependencies": {
+    "react": "^15.3.0",
+    "react-dom": "^15.3.0"
   }
 }
--- a/frontend/src/components/ExploreHeader.js
+++ b/frontend/src/components/ExploreHeader.js
@@ -1,0 +1,11 @@
+import React, { Component, PropTypes } from 'react'
+
+class ExploreHeader extends Component {
+  render = () => {
+    return (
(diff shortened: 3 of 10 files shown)


frontend/src/components/ExploreHeader.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component, PropTypes } from 'react'
```
devvmh: These imports could be changed to requirejs format if we want to

frontend/src/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+// this is optional really, if we import components directly React will be 
+// in the bundle, so we won't need a global reference
+window.React = React
+window.ReactDOM = ReactDOM
+
+window.Metamaps = window.Metamaps || {}
+window.Metamaps.ReactComponents = {
+  ExploreHeader
+}
```
devvmh: I envision either having a bunch of components available for use in the javascript files over in app/assets/javascripts, or this file could call ReactDOM.render a bunch of times.

frontend/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+    ]
+  },
+  entry: {
+    'metamaps.bundle': './src/index.js'
+  },
+  output: {
+    path: '../app/assets/javascripts/webpacked',
+    filename: '[name].js'
+  }
+}
```
devvmh: This file just basically sets things up so running `webpack` builds `src/index.js` and all imported modules into a bundle at `../app/assets/javascripts/webpacked/metamaps.bundle.js`

frontend/.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    "react",
+    "es2015"
+  ],
+  "plugins": [
+    "transform-class-properties"
+  ]
+}
```
devvmh: This file just sets up babeljs to handle react and es6 code. I was hoping to have no babel plugins (and just pure es6), but react code is much more natural to write with transform-class-properties so I bit my tongue.

without transform-class-properties, the component below would need to have a constructor and call `.bind(this)` on every function (including render) to be able to access the react props... it's just dumb. It's much simpler to use arrow functions for every function in the react component, and this works exactly how you want it to.

Connoropolous:
Beautiful Devin. What I'm not sure about is... 

would the best strategy for now be building a bunch of components, and injecting them individually into a number of different divs on the page, or would it better to write one 'app' that manages any react components we have so far?

Connoropolous:
also, to start using this code we'd need to require into application.js or directly import it into the html file yeah?

frontend/.babelrc
harlantwood: Nice.  Didn't know about `transform-class-properties`.  Thanks!

harlantwood:
Excited about this!  Great work.

frontend/package.json
```diff
@@ -16,7 +18,18 @@
...
   "devDependencies": {
+    "babel-cli": "^6.11.4",
+    "babel-loader": "^6.2.4",
+    "babel-plugin-transform-class-properties": "^6.11.5",
+    "babel-preset-es2015": "^6.9.0",
+    "babel-preset-react": "^6.11.1",
     "chai": "^3.5.0",
-    "mocha": "^2.4.5"
+    "mocha": "^2.4.5",
+    "path": "^0.12.7",
```
harlantwood: You get `path` for free from node.js, no need to put it in package.json

frontend/src/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import ExploreHeader from './components/ExploreHeader.js'
```
harlantwood: .js is optional

Connoropolous:
thanks for the input and encouragement @harlantwood :)

PR #577, opened 2016-08-01
finish integrating npm. move underscore/backbone into npm management

devvmh:
I was looking at some of the other libraries (require + lib directories) but it's hard to see exactly what they export.

I say if it ain't broke, don't fix it, for now.


Connoropolous:
I am not remembering *precisely* what it was, but I feel like there may have been a reason why I couldn't just switch to the latest version of backbone and underscore. The version numbers that they were at before may have been important a.k.a. things might be broken if a higher number installs... worth testing I think. if it is the case, we can still keep them moved over to npm, but just set hard values on the version numbers

devvmh:
Thats what i did on this pr, i set them to the current versions. But yeah
we could switch ^ to =

Connoropolous:
We just need to change underscore from 1.8.3 to 1.4.4 because that's what the static file was that's getting removed.

PR #578, opened 2016-08-01
remove all traces of Famous and improve mobile

Connoropolous:
still to do 
- [x] make sure topic view works
- [x] style the toast message
- [x] get opening and closing of chat tray displaying nicely


app/assets/javascripts/application.js
```diff
@@ -43,7 +43,5 @@
 //= require ./src/Metamaps.Mapper
 //= require ./src/Metamaps.Admin
 //= require ./src/Metamaps.Import
+//= require ./src/Metamaps.Header
 //= require ./src/Metamaps.JIT
-//= require_directory ./shims
```
Connoropolous: @devvmh so many of these things were JUST in here for famous and now we can toss them!

app/assets/javascripts/src/JIT.js.erb
```diff
@@ -2730,15 +2727,7 @@ var Canvas;
         }
       }
       this.element.appendChild(this.labelContainer);
-
-      // START METAMAPS CODE
-      var m = Metamaps.Famous.viz.surf;
-      m.setContent(this.element);
```
Connoropolous: Escape my freaky hack to make JIT work with Famous

app/assets/javascripts/src/Metamaps.GlobalUI.js
```diff
@@ -43,31 +43,45 @@ Metamaps.Active = {
...
+  // load whichever page you are on
+  if (Metamaps.currentSection === "explore") {
+      var capitalize = Metamaps.currentPage.charAt(0).toUpperCase() + Metamaps.currentPage.slice(1);
+
+      Metamaps.Views.exploreMaps.setCollection( Metamaps.Maps[capitalize] );
+      if (Metamaps.currentPage === "mapper") {
+          Metamaps.Views.exploreMaps.fetchUserThenRender();
+          Metamaps.Header.fetchUserThenChangeSection(!!Metamaps.Active.Mapper, Metamaps.Maps.Mapper.mapperId)
+      }
+      else {
```
Connoropolous: Perform the page initialization that was happening in the Famous code in the main thread, after everythings been initialized

app/assets/javascripts/src/Metamaps.Header.js
```diff
@@ -0,0 +1,22 @@
...
+    $.ajax({
+       url: '/users/' + mapperId + '.json',
+       success: function (response) {
+         Metamaps.Header.changeSection(signedIn, 'mapper', response.image, response.name)
+       },
+       error: function () {}
+    }); 
+  },
+  changeSection: function (signedIn, section, userAvatar, userName) {
+    ReactDOM.render(
```
Connoropolous: @devvmh this is where I wrap up the React rendering in a function within the relevant module

app/assets/javascripts/src/Metamaps.Views.js
```diff
@@ -80,24 +80,8 @@ Metamaps.Views.init = function () {
...
-
-      if (!Metamaps.Views.initialized) {
-        m.deploy(m._currTarget)
-        Metamaps.Views.initialized = true
-        setTimeout(updateHeight, 100)
-      } else {
-        setTimeout(updateHeight, 100)
-      }
-
+      $('#exploreMaps').empty().html(this.el)
```
Connoropolous: Making backbone views work is soooo much easier without Famous

public/famous/main.js
```diff
@@ -1,408 +0,0 @@
...
-    // import dependencies
-    var Engine = require('famous/core/Engine');
-    var Modifier = require('famous/core/Modifier');
-    var Transform = require('famous/core/Transform');
-    var Surface = require('famous/core/Surface');
-    var Timer = require('famous/utilities/Timer');
-    var Scrollview = require('famous/views/Scrollview');
-    var ContainerSurface = require('famous/surfaces/ContainerSurface');
-    var RenderNode = require('famous/core/RenderNode');
-
```
Connoropolous: This file feels like the most "JUST MAKE IT WORK" file of all time. I can't believe all of this crazy code has been running Metamaps for 1.5 years

frontend/src/index.js
```diff
@@ -1,17 +1,18 @@
...
 import Backbone from 'backbone'
 import _ from 'underscore'
-import ExploreHeader from './components/ExploreHeader'
 
-// this is optional really, if we import components directly React will be 
+// this is optional really, if we import components directly React will be
 // in the bundle, so we won't need a global reference
 window.React = React
 window.ReactDOM = ReactDOM
+Backbone.$ = window.$
```
Connoropolous: @devvmh apparently we have to do this, to keep the backbone router working

app/assets/javascripts/src/Metamaps.GlobalUI.js
```diff
@@ -1,4 +1,4 @@
-var Metamaps = {}; // this variable declaration defines a Javascript object that will contain all the variables and functions used by us, broken down into 'sub-modules' that look something like this
+var Metamaps = window.Metamaps || {}; // this variable declaration defines a Javascript object that will contain all the variables and functions used by us, broken down into 'sub-modules' that look something like this
```
Connoropolous: @devvmh based on the order that these files are getting required into application.js without this line we override the Metamaps.ReactComponents object that we created before and lose access to it.

devvmh: oh yes, awesome. makes it more portable anyways

frontend/src/index.js
devvmh: Oh brilliant, i was worried that wasn't going to work, I saw an error yesterday afternoon and was worrying about it all night lol. You're a genius

app/views/main/home.html.erb
```diff
@@ -32,11 +32,6 @@
     </div>
   </div>
 </div><!-- end yield -->
-<div class="github-fork-ribbon-wrapper right-bottom">
-  <div class="github-fork-ribbon">
-    <a href="https://github.com/metamaps/metamaps" target="_blank">Fork me on GitHub</a>
-  </div>
-</div>
```
devvmh: Not that I'm opposed to removing, but what was your reasoning? Is it because it makes it wonky when the screen is small?

frontend/package.json
```diff
@@ -28,10 +28,9 @@
     "webpack": "^1.13.1"
   },
   "dependencies": {
-    "backbone": "^1.0.0",
+    "backbone": "1.0.0",
     "react": "^15.3.0",
     "react-dom": "^15.3.0",
-    "requirejs": "^2.1.1",
-    "underscore": "^1.4.4"
+    "underscore": "1.4.4"
```
devvmh: Ah thanks for fixing backbone/require/underscore versioning. Nice

frontend/src/components/Header.js
```diff
@@ -0,0 +1,59 @@
...
+              <div className="exploreMapsIcon"></div>
+              Authorized Apps
+            </a> : null }
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
+
```
devvmh: In general, I think you could harness a few techniques to clean up:

1) I don't know if this is best practice, but I like putting null at the top in ternary, so I can indent the JSX element. I find it looks nicer. The other option is changing the rendering of the conditional element to a helper that returns null or the element
2) In general, I think you could add helper functions for a lot of the conditional code
3) Try ES6 object destructuring. It's super fun - wacky the first time you see it.
4) React has a great section about defining the types of your props, for better code documentation and verification during development. I think it's a good best practise.

I'll post a diff with some of this implemented in a second

devvmh: Do you know about var/let/const? It's really non-obvious, but it helps out with React code because when you deal with React state, you always try to avoid mutating it.

`var` behaves like the old javascript var. It's function-scoped, and if you leave it off it's a global var.
`const` creates a constant, but almost all the time in my es6/react code I use const. If it's an object or array, you can still add/delete/push/pop properties, so I use const 90% of the time. `const` is also block-scoped (so inside if/for/while, instead of being all over the function), unlike var, which makes it nicer to work with than var.

`let` is more like a traditional variable - it allows mutations. It's also block-scoped. You mostly only need this for changing numbers/strings, or occasionally if you need to replace the entire contents of an array/object. A new es6 pattern is `for (let i = 0; i < 10; i += 1) {`; you can see let replaces var.

Most ES6 people say just stop using var (as we've done here) in es6 code. The reason is that let and const are a) block-scoped and b) communicate more about the intent of the variable.

frontend/src/index.js
devvmh: This is great. There are big problems with npm-packaged jquery, and this solves that neatly

frontend/src/components/Header.js
devvmh: here are three commits you could consider for the component. I don't think any of them are mandatory. I think the third one is the most useful, but also the most complicated

https://github.com/metamaps/metamaps/compare/paring.down...paring.down.change-react

devvmh:
BTW, this PR is great. I can't believe you were able to just remove all that code. Pretty incredible

app/views/main/home.html.erb
Connoropolous: No, it's because in my new homepage mockup, I am thinking we can add a dedicated little section to the code itself.

frontend/src/components/Header.js
Connoropolous: awwwwwesome. thanks for the info and the code. no I did not know the details of const, let, var. I'm merging your commits

devvmh:
Hmm so `app/assets/images/junto_screen.jpg` has a slightly more zoomed out view than `app/assets/images/junto_screen.png`. The JPG version includes a fourth video bubble that's all in black. I'm not sure where you're using the images exactly, but perhaps the JPG should be changed to look like the PNG

Connoropolous:
Oh yeah, the one file should be deleted. All of that stuff is actually just
in there temporarily. I don't think I would use those screenshots for the
final draft of a revised homepage

Connoropolous:
@devvmh this is sooo close to be being ready, now that I've taken the homepage changes out.

devvmh:
sweet! i'll wait until it's done and review? or let me know if i should review now

Connoropolous:
One of the troubles with all of this, is that the animations that were being performed with famous, were outside of the main JS thread, which meant they happened faster. Now, with jquery animate driving the animations, it gets visibly tight in terms of the performance during the transitions... I am going to think about how to improve this...

devvmh:
Could we just pull some animations off of the main thread? It wont work for
every one but just do

```
window.setTimeout (function() {
  $.animate(...)
}, 0)
```

Or add a promise polyfill and do

```
new Promise(function(rs, rj) {
  $.animate(...)
}
```

devvmh:
Could we just pull some animations off of the main thread? It wont work for
every one but just do

```
window.setTimeout (function() {
  $.animate(...)
}, 0)
```

Or add a promise polyfill and do

```
new Promise(function(rs, rj) {
  $.animate(...)
}
```

Connoropolous:
ok, this is ready for review I think. It doesn't help that I added the mobile stuff, but you can test it all out live right now on metamaps.herokuapp.com :) 

next steps I'd like to take:
- [ ] further tiny tweaks to improve the homepage
- [ ] a separate page for the 'about metamaps/us'
- [ ] a page for creating a new map, specifically for use on mobile
- [ ] improvements to the experience of navigating around a map on mobile including
  - [ ] panning
  - [ ] usable topic card
- [ ] a dedicated search / search results page, that would be usable on mobile as well

PR #582, opened 2016-08-03
attempt to fix purple metacodes (#527)

devvmh:


devvmh:
gah i wanted master not develop

PR #583, opened 2016-08-03
attempt to fix purple metacodes (#527)

devvmh:
We should continue to test the hotfixed version on metamaps.cc before merging this.

I am a fan of moving more javascript into app/assets/javascripts and out of app/views, though, in a general sense


Connoropolous:
How/why does this work?

devvmh:
Haha, I have no idea

I found the most interesting information here http://stackoverflow.com/questions/29418465/unable-to-load-cross-origin-image-from-cloudfront-in-safari

It suggested two things:

1. The problem could be "requesting an image before my CORS setup"
2. Adding a query string can help

I tried various iterations, and this combination seemed to help. I still don't know what "my CORS setup" even means, but it twigged me in to moving the image loads earlier.

The metacode images are requested 3-4 times on the page... Sometimes it's through javascript (which is subject to CORS) and sometimes it's img tags on the page. I imagine the img tags were working but the CORS was failing, which is why you'd get purple on the map and valid icons in the metacode carousel.

Other than that... I don't really know. I do still get the sense that this is a temporary problem, and that in 1-2 years it won't be as big. But I could be wrong.

Another possibility is that requested ?purple=stupid means it doesn't "share" the images it loads with the page. That is, maybe the HTML parser loads the images once, and that was interfering with JIT loading the images. I'm not sure.

Connoropolous:
+R 

lol. I haven't always needed to know how something is working to be happy with adding it to metamaps :)
besides, we KINDA know-ish

PR #599, opened 2016-08-10
inject NODE_ENV into webpack - probably fixes #595

### LINKED ISSUE #595: properly compile react for production
`Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.`

https://facebook.github.io/react/downloads.html

http://dev.topheman.com/make-your-react-production-minified-version-with-webpack/

devvmh:


Connoropolous:
+R nice, makes sense :)

PR #603, opened 2016-08-12
enable panning on touchscreens

Connoropolous:
hey, I actually learned something about how JIT works 

:)


Connoropolous:
also, dragging nodes around also works now!

also, I was trying to get zoom working, but the code for that is commented out because I was wasting my time. Can come back to that.

devvmh:
Interesting @Connoropolous , why is it better to use JIT.js than Metamaps.JIT.js?

Connoropolous:
Because I figured I would just extend the Navigation module within JIT in
the way that it appeared to me he himself intended to. It was *almost*
there already so I just implemented it based on the existing framework that
was there for it.

Issue #580, opened 2016-08-02
make map creation more like creating a google doc: name it after it gets created


Connoropolous:
this is because when you want to get mapping, you want to get mapping. make it SO easy.

aspects of implementation: 
- map doesn't get created in the db until first topic or synapse gets added
- make it smart about prompting you to get a name and description for the map

PR #607, opened 2016-08-15
layout tweaks

Connoropolous:
summary of changes 
- added the logo in the top left, where the home button was before
- keep the search box open at all times, shrink the icon size
- hide the search box when on the logged out home page
- change map creation to be automatic, in a new tab, and to prompt for name once on that new map
    - Devin reacted with :+1:
- hide the 'mapping help' when not actually on a map

- enable the creation of a new map, by hitting a url, if you're signed in: this means that we'll be able to send people a link to create a new map, which is pretty cool
its: `/maps/new`

- using the querystring `?new` at the end of your map url will automatically open the map infobox, and also a browser prompt asking the user to enter a map title

Changes to be made soon: 
- enable searching by default when on a map, to be just on that map (like github within repos)
- improve the inviting of collaborators, and the general permission/visibility experience


Connoropolous:
using some design references here:
screen shot 2016-08-15 at 4 22 54 pm
screen shot 2016-08-15 at 4 09 57 pm

screen shot 2016-08-15 at 2 21 52 pm
screen shot 2016-08-15 at 2 21 38 pm
screen shot 2016-08-15 at 2 22 07 pm

app/assets/javascripts/src/Metamaps.Map.js
```diff
@@ -436,11 +436,25 @@ Metamaps.Map.InfoBox = {
...
     self.generateBoxHTML = Hogan.compile($('#mapInfoBoxTemplate').html())
+    
+    var querystring = window.location.search.replace(/^\?/, '')
+    if (querystring == 'new') { 
+      self.open()
+      window.setTimeout(self.requestMapTitle, 3000)
+    }
+  },
+  requestMapTitle: function () {
+    var t = prompt('Please enter a name for your map. You can also set it any time using the map info panel.')
```
devvmh: Are prompts too 90's? Should we get some kind of modal dialog library, like bootbox or something?

I'm thinking this sentence could be better..... not exactly sure. Maybe

Please enter a name for your map. You can change this later at any time using the map info box.

Box/panel? Is it clear what that means? Before the map info box could be blissfully ignored if you were new to Metamaps, but now it's a new introduced concept. Dunno. How do people do those interactive tutorials on people's first sign ins?

app/controllers/maps_controller.rb
```diff
@@ -86,6 +86,20 @@ def usermaps
     end
   end
 
+  # GET maps/new
+  def new
+    @map = Map.new(:name => "Untitled Map", :permission => "public", :arranged => true)
+    authorize @map, :create?
```
devvmh: You can just say authorize @map 

app/policies/application_policy.rb defines this function:

    def new?
      create?
    end

app/controllers/maps_controller.rb
```diff
@@ -86,6 +86,20 @@ def usermaps
     end
   end
 
+  # GET maps/new
+  def new
+    @map = Map.new(:name => "Untitled Map", :permission => "public", :arranged => true)
```
devvmh: In new ruby you can rewrite to

    @map = Map.new(name: "Untitled Map", permission: "public", arranged: true)

Hold on!! Any reason you went with public and not commons?

app/views/layouts/_upperelements.html.erb
```diff
@@ -4,7 +4,7 @@
 <div class="upperLeftUI">
   <!-- home button -->
   <div class="homeButton">
-    <a href="<%= root_url %>" <% if !current_user || appsPage %><%= 'data-bypass=true' %><% end %>></a>
+    <a href="<%= root_url %>" <% if !current_user || appsPage %><%= 'data-bypass=true' %><% end %>>METAMAPS</a>
```
devvmh: METAMAPS lol

app/controllers/maps_controller.rb
Connoropolous: Because I think it might not be right to make the default setting for maps
that people create ones in which anyone at all with an account can edit
their map. This has worked well when we have so few users, but it doesn't
bode well with more people signing on. I think we have to give them a
better clue how to
1. Invite people to collaborate on their map
2. Better display the option to convert their map to Commons, and why

Along with that, it would be great to be able to revert your Commons map
back to before someone accidentally or intentionally removed everything
from it

app/assets/javascripts/src/Metamaps.Map.js
Connoropolous: The prompt IS pretty 90s. There are a few options here and I'm open to them:

Just open the map info box, and draw attention somehow to the Map Name

Do a little custom prompt that is nicer

app/controllers/maps_controller.rb
Connoropolous: K!

Connoropolous:
closes #580

PR #609, opened 2016-08-16
fix login redirect loop

devvmh:
The loop only came up on http:// versions of metamaps (heroku staging and dev envs), and further only if you logged in from /login (much more likely on the mobile view, since it pops up in a box on the homepage in desktop view).

This fixes it in any case


Connoropolous:
+R groovy

PR #615, opened 2016-08-21
Enable users to star maps, and to see their starred collection

Connoropolous:
needs design, just a 'star/unstar' icon / UX

screen shot 2016-08-20 at 9 11 06 pm
screen shot 2016-08-20 at 9 11 13 pm
screen shot 2016-08-20 at 9 11 21 pm


app/assets/javascripts/src/Metamaps.Map.js
```diff
@@ -49,12 +49,18 @@ Metamaps.Map = {
       return false
     })
 
+    $('.starMap').click(function () {
+      if ($(this).text() === 'star') self.star()
+      else self.unstar()
```
devvmh: :'( I might look into if this can be done more nicely with react. I'll need to figure out if the state needs to be in sync with Backbone, or just with the server. It's more annoying if it needs to sync with backbone

Connoropolous: Lol, yeah this is super hacky. I coded this whole feature in two hours on Saturday night, and rushed a few little things, knowing that there'd be a few changes to make when I  get an image asset for the starring/unstarring. I will do something to improve it at that point :)

PR #618, opened 2016-08-24
[ON HOLD] list maps that are live with someone on the map

Connoropolous:
@poietic I have been thinking... we could rename the 'recently active' collection to "global" which is a better explanation for the pool of maps that it is displaying, rather than how it's being sorted. Then, we could change the icon to a globe, and reuse the 'electric pulse' icon for this new category of 'live' maps

screenshot 2016-08-23 at 9 40 04 pm

### DIFF
--- a/app/controllers/maps_controller.rb
+++ b/app/controllers/maps_controller.rb
@@ -71,7 +71,7 @@ class MapsController < ApplicationController
       end
     end
   end
-
+  
   # DELETE maps/:id
   def destroy
     @map.delete
--- a/app/assets/stylesheets/clean.css.erb
+++ b/app/assets/stylesheets/clean.css.erb
@@ -751,6 +751,10 @@
     background-image: url(<%= asset_data_uri 'exploremaps_sprite.png' %>);
     background-position: 0 0;
 }
+.exploreMapsCenter .liveMaps .exploreMapsIcon {
+    background-image: url(<%= asset_data_uri 'exploremaps_sprite.png' %>);
+    background-position: -32px 0;
+}
 .exploreMapsCenter .featuredMaps .exploreMapsIcon {
     background-image: url(<%= asset_data_uri 'exploremaps_sprite.png' %>);
     background-position: -96px 0;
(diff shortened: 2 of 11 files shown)


devvmh:
What about "Browse global maps" or someting like that

And maybe "Starred by me"

Connoropolous:
Yeah, I think you're right, adding a bit more clarification. Maybe 'Global Maps' and 'Starred By Me'

devvmh:
did a merge commit from develop

app/controllers/maps_controller.rb
```diff
@@ -72,6 +72,14 @@ def sharedmaps
...
     end
   end
+  
+   # GET /explore/live
+  def livemaps
+    skip_policy_scope
+    respond_to do |format|
+      format.html { respond_with() }
+    end
+  end
```
devvmh: I'm assuming this will be policy scoped at some point right?

app/controllers/maps_controller.rb
```diff
@@ -1,6 +1,6 @@
 class MapsController < ApplicationController
   before_action :require_user, only: [:create, :update, :access, :star, :unstar, :screenshot, :events, :destroy]
-  after_action :verify_authorized, except: [:activemaps, :featuredmaps, :mymaps, :sharedmaps, :starredmaps, :usermaps]
+  after_action :verify_authorized, except: [:activemaps, :featuredmaps, :mymaps, :sharedmaps, :starredmaps, :usermaps, :livemaps, :events]
   after_action :verify_policy_scoped, only: [:activemaps, :featuredmaps, :mymaps, :sharedmaps, :starredmaps, :usermaps]
```
devvmh: This section is getting pretty long. What if we had a MapsIndexController just for these different index methods?

devvmh: Another goal of mine is to have verify_authorized happen on every controller method in the whole app. So all of these exceptions would eventually disappear, leaving just `after_action :verify_authorized`. That mostly means making policies with a bunch of functions that return true

Connoropolous:
closing in favor of #775

PR #621, opened 2016-09-01
proposal: automatic versioning via git

devvmh:
@Bortseb does this work on Windows? Are you able to fetch this branch and try to run the server to see if it works?


Connoropolous:
let's just proceed with this, and do a fix for windows if it ever gets raised as an issue

PR #626, opened 2016-09-02
shouldn't reference relatives that are connected in private ways

### LINKED ISSUE #346: unattached topic within topic view - unidentified bug,
unattached topic within topic view - unidentified bug, see here: http://metamaps.cc/topics/10472, http://metamaps.cc/topics/12190

Connoropolous:
closes #346 

@devvmh could you consult with me on this? how could we make the code more DRY? 

especially since this code is identical to the code in the Synapse policy_scope

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -40,7 +40,7 @@ class TopicsController < ApplicationController
     @topic = Topic.find(params[:id])
     authorize @topic
 
-    @alltopics = [@topic].concat(policy_scope(Topic.relatives1(@topic.id, current_user)).to_a).concat(policy_scope(Topic.relatives2(@topic.id, current_user)).to_a)
+    @alltopics = [@topic].concat(policy_scope(Topic.relatives(@topic.id, current_user)).to_a)
     @allsynapses = policy_scope(Synapse.for_topic(@topic.id))
 
     @allcreators = @alltopics.map(&:user).uniq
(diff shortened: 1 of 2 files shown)


devvmh:
OK i posted a PR that... I think will fix it. You'd better review

Why is there a relatives1 and relatives2 scope anyways? Was that there all along? I'm not clear exactly on how the logic works there. Could we clean up those functions while we're mucking around in here?

Otherwise LGTM merge wheneva

Connoropolous:
There's relatives1 and relatives2 because I didn't know how else to get all
the relatives, while also having all of them be policy_scoped. Would love
to see it happen cleaner. The code that calls this stuff that is important
is found in the show and network actions of the topic controlller

devvmh:
Hmm so maybe it would be

```
scope :relatives, ->(topic_id = nil, user = nil) {
  synapses = Pundit.policy_scope(user, Synapse.where(node1_id: topic_id)).map(&:node1_id)
  synapses += Pundit.policy_scope(user, Synapse.where(node2_id: topic_id)).pluck(&:node2_id)
  where(id: synapses.uniq)
}
```

Connoropolous:
Yes, I think your last comment is on point

Connoropolous:
Except that I think the .map and .pluck have to pull the inverse node id than the one they're querying for

app/controllers/topics_controller.rb
```diff
@@ -66,7 +64,7 @@ def relative_numbers
 
     topicsAlreadyHas = params[:network] ? params[:network].split(',').map(&:to_i) : []
 
-    @alltopics = policy_scope(Topic.relatives1(@topic.id)).to_a.concat(policy_scope(Topic.relatives2(@topic.id)).to_a).uniq
```
devvmh: lines 69 and 91 used to have .uniq, but now that's inside the `relatives` scope.

Connoropolous:
Yees :) fantastic Devin

PR #628, opened 2016-09-03
use policy scope in bug/detached

devvmh:
Update for #626 - the real trick is using the `Pundit.policy_scope(user, scope)` function.

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -43,33 +43,13 @@ class Topic < ActiveRecord::Base
   end
 
   scope :relatives1, ->(topic_id = nil, user = nil) {
-     visible = %w(public commons)
-     permission = 'synapses.permission IN (?)'
-     if user
-       synapse_permission = permission + ' OR synapses.defer_to_map_id IN (?) OR synapses.user_id = ?' 
-       return includes(:topics1)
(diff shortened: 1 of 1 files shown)


app/models/topic.rb
```diff
@@ -43,33 +43,13 @@ def relatives
...
-       synapse_permission = permission + ' OR synapses.defer_to_map_id IN (?) OR synapses.user_id = ?' 
-       return includes(:topics2)
-         .where('synapses.node2_id = ? AND (' + synapse_permission + ')', topic_id, visible, user.shared_maps.map(&:id), user.id)
-         .references(:synapses)
-     else
-       return includes(:topics2)
-         .where('synapses.node2_id = ? AND (' + permission + ')', topic_id, visible)
-         .references(:synapses)
-     end
+    synapses = Pundit.policy_scope(user, Synapse.where(node2_id: topic_id))
```
Connoropolous: How does Pundit.policy_scope know to use the synapses policy?

devvmh: because the Scope is a Synapse class, so it grabs that name and adds 'Policy' on the end and looks for a policy with that name

You could also replace it with `SynapsePolicy::Scope.new(user, ...)` I think.

It all uses this `find` method under the hood: https://github.com/elabs/pundit/blob/master/lib/pundit/policy_finder.rb#L76

devvmh:
Committing directly to bug/detached branch, see other PR

PR #633, opened 2016-09-09
use a subset of the simplest metacodes for new users

Connoropolous:
this is because 
A: duh
B: this piece of user feedback https://metamaps.slack.com/messages/ux-feedback/files/F29V6KNFP/

The thinking behind why I chose the ones that I did choose are because they are the most straightforward ones that require no deep reflection to understand and use. 
The list of 14 is: 
Con Event Group Idea Insight Intention Need Opportunity Person Pro Problem Project Question Wildcard

so new users start with this: 
screenshot 2016-09-09 at 3 47 20 pm

instead of this: 
screenshot 2016-09-09 at 3 48 06 pm

### DIFF
--- a/app/models/user_preference.rb
+++ b/app/models/user_preference.rb
@@ -3,7 +3,7 @@ class UserPreference
 
   def initialize
     array = []
-    %w(Action, Aim, Idea, Question, Note, Wildcard, Subject).each do |m|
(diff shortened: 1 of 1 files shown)


Connoropolous:
ok final list for V1 is action, aim, idea, question, note, wildcard, subject

harlantwood:
Excellentl! Was super overwhelming for me too.

PR #639, opened 2016-09-13
Allow pasting URLs into the map to create a new topic

### LINKED ISSUE #638: paste url into a map creates a link/url topic
Then you can drag bookmarks or whatever onto a map

mind=blown

devvmh:
Also includes:

- drag and drop files onto the map (json, tsv, .webloc, or txt with a url in it)
- parse OSX .webloc files for URLs (drag a bookmark onto your desktop, then drag it from Desktop to metamaps and bam! new topic)
- Extract PasteInput object from the Import object to keep Import focussed on import
- AutoLayout object to provide x & y values for topics with no x/y in a spiral

fixes #638 

I think I've decided xls/xlsx/ods input is too much. we don't need it. I'll set up CSV input because duh, but probably in another PR because my wonky csv format is bad

If we merge this, it makes setting up csv import more urgent, but also more doable. We can just drag the file onto the canvas now.

It's worth asking if this UI is too much; perhaps we should drop back to a safer file upload field.

- [x] switch to the current auto layout that goes in a "square spiral" and delete this new mathematical spiral code

optional TODO:

- [ ] set up csv import

### DIFF
--- a/app/assets/javascripts/src/Metamaps.Import.js
+++ b/app/assets/javascripts/src/Metamaps.Import.js
@@ -254,7 +254,7 @@ Metamaps.Import = {
   },
 
   createTopicWithParameters: function (name, metacode_name, permission, desc,
-    link, xloc, yloc, import_id) {
+    link, xloc, yloc, import_id, opts) {
     var self = Metamaps.Import
     $(document).trigger(Metamaps.Map.events.editedByActiveMapper)
     var metacode = Metamaps.Metacodes.where({name: metacode_name})[0] || null
@@ -271,7 +271,8 @@ Metamaps.Import = {
       permission: topic_permission,
       defer_to_map_id: defer_to_map_id,
       desc: desc || "",
-      link: link || ""
+      link: link || "",
+      calculated_permission: Metamaps.Active.Map.get('permission')
     })
     Metamaps.Topics.add(topic)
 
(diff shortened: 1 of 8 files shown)


devvmh:
This could be merged immediately, or could be improved a bit. I could have added WIP to the title but didn't....

app/assets/javascripts/src/Metamaps.Import.js
```diff
@@ -272,15 +263,22 @@ Metamaps.Import = {
...
       name: name,
       metacode_id: metacode.id,
-      permission: permission || Metamaps.Active.Map.get('permission'),
-      desc: desc || "",
-      link: link
+      permission: topic_permission,
+      defer_to_map_id: defer_to_map_id,
+      desc: desc || ""
     })
+    topic.set('desc', desc || '') // TODO why is this necessary?
```
Connoropolous: these two overrides are required because Metamaps.Backbone.Topic has an init function which is currently overriding passed in values for Desc and Link. Going into Metamaps.Backbone.js and editing the init function for Topic will solve it

devvmh:
Probably once this is ready and working, we could phase out pasting input. It's much nicer to drag/drop files on.

Connoropolous:
looks good, going in?

devvmh:
Yeah, sure!

PR #640, opened 2016-09-14
fix password reset error

devvmh:
UNTESTED - test on localhost and/or heroku before merging!!


devvmh:
doesn't work on localhost/heroku because the mailer isn't setup, i'll try hotfixing production and test there

devvmh:
hmmm it seems to work with/without the hotfix 😢

devvmh:
I think this is more helpful than not, so I'll merge

PR #641, opened 2016-09-17
Get siblings by metacode type returns only topics with that metacode - fix #538

### LINKED ISSUE #538: [TOPIC VIEW] - Get Siblings by metacode type returns *all* siblings instead of selected type
Selecting a metacode type from the context menu to add a topic's siblings to the canvas in topic view does not limit to the selected type, instead adding all siblings of the selected topic into the view.

devvmh:


Connoropolous:
Sweet! I wonder how this code ended up without that in it, when originally it did work.

devvmh:
I looked a bit through the log today... I didn't find it but I suspect it was me during the upgrade to rails 4.

PR #643, opened 2016-09-17
allow topic carousel to be pinned open

Connoropolous:
so when you pin the carousel open, topics can be repeatedly created, and they are created using auto-placement, starting at the "origin" the center of the map. If the auto-layout has been used to create topics previously, it will recursively look for the first "open position" at which to place the new topic. also, the topic carousel stays open, even when you click on an empty piece of the canvas, unlike when it is not pinned, and closes

also includes a bonus tooltip above the 'switch metacodes' gear

screen shot 2016-09-17 at 7 02 55 pm
screenshot-metamapsdotcc-connoropolous c9users io 2016-09-17 19-05-57


devvmh:
Oh man this is so much better than my approach. I didn't realize this code existed, because it's exactly what I imagined, but I was trying to use math to save code

I'll delete the spiral code from my code and start using this one.

devvmh:
LGTM for merging

PR #648, opened 2016-09-22
hack to load link titles when pulling a url into a map

devvmh:


devvmh:
OK, I added the rack-attack gem for rate limiting the `hacks/load_url_title` route, so I'm feeling a bit better about the security of this. 

It is a bit scary that someone can load arbitrary urls. Luckily they have to be signed in. This is definitely a route we'll want to consider and reconsider in the future.

On the bright side, rack-attack will help us implement rate limiting for our API in the future if we want! I couldn't figure it out tonight but it'll be doable once we are ready for it.

devvmh:
ok fine i added what i think is rate limit headers. (they might only show up once you've hit the limit, it's unclear)

PR #655, opened 2016-09-23
Feature/fix map titles

Bortseb:
This little change makes it so that maps launched from the explore maps page shows the map name in the browser tab, instead of the map id.


Bortseb:
Not really sure what that is all about, backbone models, maybe explain one day when there is time... But i'll change it now to be consistent.

Connoropolous:
Backbonejs.org would explain :)

Connoropolous:
So we usually leave it to the person made the PR to merge it, cause it's satisfying. It's been approved now, so feel free to merge it in!

PR #659, opened 2016-09-24
simplify js

devvmh:
most of the js left in app/assets/javascripts uses jQuery, so it's hard to do

It would be great to figure out how to get jQuery via npm


harlantwood:
jquery via npm works well... Any issues you've had there?

devvmh:
I tried moving jquery out of rails into npm and exposing it on window, but there were a bunch of issues I didn't totally understand

If you can get it working in a PR that would be so great

devvmh:
I originally had pulled countable and howler out of app/assets/javascripts and into npm in this PR, but it seems to cause problems with our testing. For now I'm leaving that on a branch and merging these more useful changes first.

PR #676, opened 2016-09-27
delete old gems and upgrade aws/paperclip

### LINKED ISSUE #657: AWS SDK v1 has been deprecated in paperclip 5.
We currently use paperclip 4

- [ ] upgrade to AWS SDK v2, including checking syntax/breaking changes and testing functionality
- [ ] upgrade to paperclip 5, including checking syntax/breaking changes, and testing functionality

devvmh:
This needs serious testing!! It might make sense to wait until after Oct 31 to merge.

Fixes #657 

What did I do:

1. update paperclip & aws
2. switch map screenshots to be uploaded as a blob instead of base64, which saves on bandwidth and amount of code
3. remove old gems, including rails12factor, and update config to emulate their behaviour
4. Update doorkeeper too, for some reason. I don't think this changed anything.

### DIFF
--- a/app/controllers/maps_controller.rb
+++ b/app/controllers/maps_controller.rb
@@ -1,12 +1,9 @@
 # frozen_string_literal: true
 class MapsController < ApplicationController
-  before_action :require_user, only: [:create, :update, :destroy, :access, :events,
-                                      :screenshot]
+  before_action :require_user, only: [:create, :update, :destroy, :access, :events]
   before_action :set_map, only: [:show, :update, :destroy, :access, :contains,
-                                 :events, :export, :screenshot]
+                                 :events, :export]
   after_action :verify_authorized
-
-  autocomplete :map, :name, full: true, extra_data: [:user_id]
 
   # GET maps/:id
   def show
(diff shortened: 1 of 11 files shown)


app/controllers/maps_controller.rb
```diff
@@ -6,8 +6,6 @@ class MapsController < ApplicationController
 
   respond_to :html, :json, :csv
 
-  autocomplete :map, :name, full: true, extra_data: [:user_id]
```
Connoropolous: why's this being removed?

Connoropolous:
ok, I  will try and test locally, so we don't have to put it in develop yet

app/controllers/maps_controller.rb
devvmh: as far as I can tell, this creates a controller action autocomplete_map_name, which isn't used anywhere else in our app, and is part of the rails3-jquery-autocomplete. I could be wrong, but I think we have different actions and are using typeahead for all of our autocomplete now. 

I definitely put this up for review in case you know of cases why this is needed. Do we still need this line/gem?

devvmh:
Cool. I have a question for you about that autocomplete line, and I can also test myself some time

... I have a one week holiday next week, so some coding will be going down at least one of the days

devvmh:
@Connoropolous I've tested this on localhost and heroku and it's working. I've updated heroku's config with an S3_REGION variable, as well as updating the .env on the production server. This is ready to merge any time. Can you give it one more review, and if you're happy, then merge it?

app/controllers/maps_controller.rb
Connoropolous: ok, yeah, it's fine for this to come out. it's not used anywhere

Connoropolous:
looooks goood!

Issue #658, opened 2016-09-24
Chrome's default context menu isn't prevented from opening when right-clicking on topics/synapses


Bortseb:
I have had this problem for a long while now. When I right click on topics/synapses, both context menus come up. The default chrome one is on top of our custom context menu.

This really gets in my way all the time, and breaks the flow of mapping. (I need to hit esc to remove the default context menu, and see our custom context menu)

I have spent a bit of time trying to solve this issue myself... But now I need some help.

I found this from research:
http://stackoverflow.com/questions/381795/how-to-disable-right-click-context-menu-in-javascript

and I found two places in our code that appear to be attempting to prevent the default context menu from coming up... But its not working it seems (Maybe just for me? Windows 10/Chrome... I haven't heard others mention this problem)

https://github.com/metamaps/metamaps/blob/eef8a281cb53af7ff333f41a6af68241e27c22f2/frontend/src/Metamaps/JIT.js#L1292

https://github.com/metamaps/metamaps/blob/develop/frontend/src/Metamaps/Map/index.js#L42

Maybe @devvmh or @Connoropolous can provide some clues?

devvmh:
I've looked at this before myself, especially after trying to map on Windows. I couldn't figure it out. I can definitely join in taking another look

Bortseb:
Based on this
http://stackoverflow.com/questions/737022/how-do-i-disable-right-click-on-my-web-page

I was trying to find a way to add `oncontextmenu="return false"` onto the canvas element with id infovis-canvas ... but not sure if I did it right to see if this would be a solution.

devvmh:
Yeah i'll bet if you updated that code on L42 it would work. I can't really test this since the deprecated code works, but if you get it working on Windows on a branch, we can just merge it to develop and then verify it's all good on different OSes by using Herokuapp

(heroku gets updated every time develop gets updated)

Bortseb:
What is confusing though is that it only happens when you right click on
topics or synapses... It doesn't happen when you right click on a blank
region of the canvas... Nothing's happens then.

devvmh:
Oh ho, that is interesting. I wonder if we can change the line from 

```
    $('#center-container').bind('contextmenu', function (e) {
      return false
    })
```

to

```
    $('#center-container div').bind('contextmenu', function (e) {
      return false
    })
```

That is, override the contextmenu event on all child divs, not just `#center-container` itself. Or even `$('#center-container *')`, although that might have some unintended consequences too

[...]

PR #678, opened 2016-09-28
Prevent default context menu from opening when topics/synapses are right-clicked

Bortseb:
Prevents the default chrome context menu from appearing overtop the Metamaps context menu

This addresses issue #658

### DIFF
--- a/frontend/src/Metamaps/Map/index.js
+++ b/frontend/src/Metamaps/Map/index.js
@@ -39,7 +39,7 @@ const Map = {
     var self = Map
 
     // prevent right clicks on the main canvas, so as to not get in the way of our right clicks
-    $('#center-container').bind('contextmenu', function (e) {
+    $('#wrapper').on('contextmenu', function (e) {
       return false
     })
 


devvmh:
Sweet! Does it work?

Bortseb:
works for me on my system... although, it does prevent more things from being right clicked than just the intended nodes/synapses... (as you predicted)... Not sure how detrimental this is though...

I tried for a while to get it down to just the topics/synapses... but something about these things being on the canvas makes this extra hard... Topics/synapses don't have the own DOM element right?

Also, for some reason selecting the canvas with jquery is also more difficult... I wish I had a good way of sourcing what event was the reason for the context menu appearing... 

It may just be that events that happen on the canvas get attributed to a parent element? and therefore a more nuanced selection isn't possible?

I dunno, I'm tired, and this does work :P

devvmh:
Cool! Then go ahead and merge it and we can try it on heroku. If I'm free tomorrow morning I might try it on localhost and post any ideas I have too

devvmh:
There is a lot of stuff that goes on with this kind of event. If you put a `debugger` statement in the callback function before `return false`, and then right click, you can inspect the state of the `e` object. It should have a `target` and a `currentTarget`. jQuery `return false` runs `e.preventDefault(); e.stopPropagation()`. If you don't stopPropagation, it'll "bubble up" to the parent elements until it hits the highest/root html tag on the page or another event handler stops the propagation

So perhaps there is a smarter way by using that `e` object, but we can always get that in another PR

Connoropolous:
heyo. So I'm going to provide an afterthought here, because I think I've actually recalled the true issue. 

on other tools, when you right click on the canvas, only our desired right click menu comes up. but if you right click on our right click menu, THEN you get the chrome right click menu. I think what is happening on windows is it is generating the other right click div which overlays, but then the click event is bubbling up and seeing a click event happening on that new menu that wasn't even there before. So if you look at the class or id of the right click menu, then you could perhaps just disable right clicking on that particular div. 
because when I was researching about this, the widespread agreement was that it is pretty bad practice to disable the browser right click menu ( so avoid it whenever possible)

Connoropolous:
so could you try changing the line instead to this and see if that works? 

$('#center-container, .rightClickMenu').bind('contextmenu', function (e) { ... })

(I don't know whether that selector actually works for the right click menu or not, you'd have to find out)

PR #679, opened 2016-09-28
fix develop branch bugs

devvmh:


app/controllers/search_controller.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-class MainController < ApplicationController
+class SearchController < ApplicationController
```
devvmh: @connoropolous here's that heroku bug fixed

Connoropolous: ok, sweet. got it

PR #681, opened 2016-09-28
Better default context menu suppression fix

Bortseb:
This is a much better context menu fix than the one I submitted last night. @Connoropolous was right that it was the div.rightclickmenu element that was triggering the unwanted context menu (When I looked at the target in the event object)...

However trying to add an event handler in the index.js file, as I was solving this problem initially, didn't work... I figured, maybe this contextmenu event was being triggered before I even had the chance to to add the event handler... So I went right back to the exact moment in the code where the div.rightclickmenu was created, and added the oncontextmenu="return false" attribute right afterwards...

This has solved the problem in my c9.io instance, and I'm no longer disabling the context menu on every div element either :P

### DIFF
--- a/frontend/src/Metamaps/Map/index.js
+++ b/frontend/src/Metamaps/Map/index.js
@@ -38,6 +38,11 @@ const Map = {
   init: function () {
     var self = Map
 
+    // prevent right clicks on the main canvas, so as to not get in the way of our right clicks
+    $('#wrapper').on('contextmenu', function (e) {
+      return false
+    })
+
     $('.starMap').click(function () {
       if ($(this).is('.starred')) self.unstar()
       else self.star()
--- a/frontend/src/Metamaps/Map/index.js
+++ b/frontend/src/Metamaps/Map/index.js
@@ -38,11 +38,6 @@ const Map = {
   init: function () {
     var self = Map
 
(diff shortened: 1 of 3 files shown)


Bortseb:
When I test the app though, it is not stopping me from right clicking and dragging to create synapses @Connoropolous

frontend/src/Metamaps/Map/index.js
```diff
@@ -38,6 +38,11 @@ const Map = {
   init: function () {
     var self = Map
 
+    // prevent right clicks on the main canvas, so as to not get in the way of our right clicks
+    $('#wrapper').on('contextmenu', function (e) {
```
Connoropolous: good, but we actually want to revert it to '#center-container' instead of '#wrapper'

Bortseb: as far as I can tell though, that id doesn't exist anymore... and when I just tested it, it doesn't prevent the default contextmenu

Connoropolous: oho. good point. the canvas element is now '#infovis-canvas'  ... try that

Connoropolous:
just as a note 'for the record', this is not ideal to be disabling the context menu for every user on these elements, but the behaviour of the right click menu coming up each time for windows user is a greater detriment, and that's what this PR fixes

PR #691, opened 2016-09-29
I changed how zoom by mouse-wheel works so that it zooms based on whe…

Bortseb:
…re your mouse pointer is


devvmh:
Cool!

is there no way to directly call Metamaps.Util from this context? It might be nice for keeping the footprint in patched/JIT.js low and for code reuse

Connoropolous:
if you think that would be better @devvmh I trust your judgment. We can import the function from util and use it

PR #707, opened 2016-10-02
finish removing global Metamaps references from frontend code

### LINKED ISSUE #706: Pass Metamaps.Erb as an argument to each init function
In `frontend/src/Metamaps/index.js`. This will remove a global reference and making testing a bit easier. It may involve moving some variable declarations inside init functions.

devvmh:
I'll merge this on Nov 2 after we do the big 3.0 deploy

Also fixes #706


Connoropolous:
@devvmh I've taken a look through at this stage, but it sounds like you're going further with it :) sweet

frontend/src/Metamaps/Control.js
```diff
@@ -110,7 +101,7 @@ const Control = {
       var mappableid = topic.id
       var mapping = node.getData('mapping')
       topic.destroy()
-      Metamaps.Mappings.remove(mapping)
+      DataModel.Mappings.remove(mapping)
```
Connoropolous: So DataModel stores data now too?

Connoropolous:
Jiminy @devvmh this is huge! So NO more Metamaps globals?

devvmh:
Yeah, almost! There are some textual onclick handlers still.

But btw this branch is hopelessly broken. There's a number of circular
dependencies,  at least one of which means the page can't initialize

devvmh:
Latest push seems to mostly work, with at least 1-2 bugs but nothing preventing a merge to develop

devvmh:
Update Oct 20: I just rebased this again. I'll continue to rebase. Assuming we release on Oct 31, I want to wait until Nov 2 then merge this, and then merge an eslint branch right away after that. Nov 1 can be for bugfixes to master.

app/views/maps/show.html.erb
```diff
@@ -9,13 +9,14 @@
...
-    Metamaps.Stars = <%= @allstars.to_json.html_safe %>;
-    Metamaps.Visualize.type = "ForceDirected";
+  Metamaps.ServerData = Metamaps.ServerData || {}
+  Metamaps.ServerData.ActiveMap = <%= @map.to_json.html_safe %>;
+  Metamaps.ServerData.Mappers = <%= @allmappers.to_json.html_safe %>;
+  Metamaps.ServerData.Collaborators = <%= @allcollaborators.to_json.html_safe %>;
+  Metamaps.ServerData.Topics = <%= @alltopics.to_json(user: current_user).html_safe %>;
+  Metamaps.ServerData.Synapses = <%= @allsynapses.to_json.html_safe %>;
+  Metamaps.ServerData.Mappings = <%= @allmappings.to_json.html_safe %>;
+  Metamaps.Messages = <%= @allmessages.to_json.html_safe %>;
```
Connoropolous: why not messages and stars under serverdata?

app/views/topics/show.html.erb
```diff
@@ -1,27 +1,20 @@
...
-    Metamaps.Synapses = <%= @allsynapses.to_json.html_safe %>;
-    Metamaps.Visualize.type = "RGraph";
+  Metamaps.currentSection = "topic"
+  Metamaps.currentPage = <%= @topic.id.to_s %>
+  Metamaps.ServerData = Metamaps.ServerData || {}
+  Metamaps.ServerData.ActiveTopic = <%= @topic(user: current_user).to_json.html_safe %>
+  Metamaps.ServerData.Creators = <%= @allcreators.to_json.html_safe %>
+  Metamaps.ServerData.Topics = <%= @alltopics(user: current_user).to_json.html_safe %>
+  Metamaps.ServerData.Synapses = <%= @allsynapses.to_json.html_safe %>
+  Metamaps.Visualize.type = "RGraph"
```
Connoropolous: This is different than in the case of ForceDirected where you changed it to `Metamaps.ServerData.VisualizeType`

frontend/src/Metamaps/DataModel/Map.js
```diff
@@ -0,0 +1,131 @@
...
+    if (mapper && this.get('user_id') === mapper.get('id')) {
+      return true
+    } else {
+      return false
+    }
+  },
+  getUser: function () {
+    return Mapper.get(this.get('user_id'))
+  },
+  fetchContained: function () {
```
Connoropolous: I took out `fetchContained, getTopics, getSynapses, getMappings, getMappers` in `develop` branch, so I don't think they should/need to be in here any more

frontend/src/Metamaps/DataModel/Mapping.js
```diff
@@ -0,0 +1,37 @@
...
+  initialize: function () {
+    if (this.isNew()) {
+      this.set({
+        'user_id': Active.Mapper.id,
+        'map_id': Active.Map ? Active.Map.id : null
+      })
+    }
+  },
+  getMap: function () {
+    return Map.get(this.get('map_id'))
```
Connoropolous: I should convert these to promises, or callback style, if we want to keep these accessor functions for relations between models.

app/views/maps/show.html.erb
devvmh: Great question, I had to look at this. 

I guess it's because they aren't backbone models, they're just arrays (like Metamaps.Metacodes)

So I could stick them in ServerData, but I'd just have to pull them back out later. Not sure what's best yet but I think this code is OK for now

app/views/topics/show.html.erb
devvmh: brilliant I'll fix this thanks

frontend/src/Metamaps/DataModel/Map.js
devvmh: thanks, rebase error, fixing now.

There may be other rebase errors like this :|

frontend/src/Metamaps/DataModel/Mapping.js
devvmh: great!

One thing about es6 promises: they need a polyfill. I considered polyfills, but eventually realized jQuery has promises built in and we're already using them. See Import.js for how I used them (with $.Deferred()). It's a bit clunkier than es6 promises, but it's not really that bad for our simple use cases.

PR #708, opened 2016-10-02
Enable access to Most Used and Recently Used metacodes in lists and carousel

### LINKED ISSUE #647: Add 'Recent' and 'Most Used' metacodes to metacode select list dropdowns
as seen in #644

Connoropolous:
closes #647 

screen shot 2016-10-02 at 4 53 28 pm
screen shot 2016-10-02 at 4 53 06 pm
screen shot 2016-10-02 at 4 52 42 pm

@devvmh It is definitely clear that this stuff related to how we save a users preferences around metacodes in the database is getting messy. This adds a teeny bit more complexity to it, but I think it is worth it for now to take this approach. It may get refactored as I do more work on the metacode carousel and that whole UX, like in my other branch that I've been working on.


Connoropolous:
wow, looks like there are a wholllle lot of style improvements that I could make, based on codeclimates recommendations. I will do those before merging...

devvmh:
I think it's more important to be consistent than to follow what code climate says. They base their style analysis on eslint, and rubocop, both of which we can set config values for in our repo. Don't let style hold you back from coding like a maniac; it's really not a very important priority

The main reason I like it is that consistency makes it easier to read and sometimes it will help me learn about a good coding practise. But spending too much time worrying about style while coding isn't helpful to productivity.

So TLDR: have fun with the style, don't let it slow you down, and we can set project-wide coding standards using rubocop and eslint config.

devvmh:
also if you're OK with the changes it suggests, you can run `rubocop -a` to apply most fixes automatically. `standard-format -w filename.js` does something similar

Connoropolous:
nice :) ok, that is helpful to have your thoughts on that

Bortseb:
I dont know what fun set is... but I think the 4 options you have are all fine to keep... Especially if fun stuff is just "your current set" (that is on the spinner too)... maybe even name it that. instead of having the set name in the list (i.e., only show 1 "set", that being your currently selected one)

Issue #701, opened 2016-10-01
Bugs found while naming newly created map... and other comments


Bortseb:
When I create a new map, the map card pops up with a nice tooltip telling me to "Click here to name the map"

I have a bunch of comments about the changes that have happened here... and some bugs to point out too.

It surprised me that clicking the new map button now immediately creates the new map. I am a little worried that this might mean that new users never learn about our map permission types because of this, or never change the type from the current default "Public". (Or never add a description)

The bugs with how things work now are:
1. When I click the topic card to start entering my name, the tooltip doesn't disappear.
2. When I click to start entering my name, the previous text "Untitled Map" isn't automatically selected to be overwritten. (Or just deleted for me)

This isn't a bug but, it might also be nice to just put the cursor into the inbox for the user, so they can just start typing... even if the tooltip is kept to grab their attention.

Connoropolous:
" I am a little worried that this might mean that new users never learn about our map permission types because of this, or never change the type from the current default "Public". (Or never add a description)" 
very valid concerns, I think you're totally right. this is an instance of development racing ahead of design. I'm not saying it's optimal, I'm saying it's something that's been happening because it feels better to have development happening regardless of whether design has been, so that things can keep moving forward. Once there is capacity within the network, or outside of it, to be thinking deeply about user experience and creating designs, then developers will be able to prioritize those UX and design changes. 

In the meantime, we are just barely holding on in terms of UX, around something like this new map create flow

Connoropolous:
in the meantime, we could make some of the simple changes that you proposed, because it seems like they will benefit the user and cause no harm

Connoropolous:
Also on this note... 
"It surprised me that clicking the new map button now immediately creates the new map. I am a little worried that this might mean that new users never learn about our map permission types because of this, or never change the type from the current default "Public". (Or never add a description)"

the reason I made this change is because I am less worried about those second step things not happening, than new users not even creating a map because they were inhibited by the initial popup

Bortseb:
I'm not trying to be critical of anyone in any way... Just added this so
that its on our list somewhere...

I don't really know what you mean by "inhibited by the initial popup" ...
You mean people would hit that page and be more reluctant to start a map?

Will the tool tip stay up even if the user starts mapping instead of first
naming their map? (if not  we could wind up with a lot of untitled maps...
Maybe there is a better default namr? So then, even if the user doesn't
choose a name...  It is still given a unique name... )

devvmh:
[...]

PR #710, opened 2016-10-02
new map improvements

Connoropolous:
- Hide maps with "Untitled Map" from the global view by default
- Remove the ?new from the url after the map info box is closed for the first time, so that refreshing the page won't pop it up again.
- When ?new is in the url, focus and select the text of Untitled Map after popping up the info box to make it easier to immediately add a title.

fixes most, but not quite everything, in #701


devvmh:
related to #701

PR #716, opened 2016-10-04
users and metacodes api endpoints

devvmh:


app/policies/user_policy.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  # API action
+  def current?
+    user == record
+  end
+
+  class Scope < Scope
+    def resolve
+      return scope.all if user.present?
+      scope.none
```
devvmh: Not sure if logged out people should be able to see users. I suppose they can get at them via javascript anyways if they can see a commons map

Connoropolous: what is it? username, id, avatar? I think it's fine if unauthed can see them

devvmh: ok will update

PR #717, opened 2016-10-04
eslint updates

devvmh:


.eslintrc.js
```diff
@@ -11,5 +11,8 @@ module.exports = {
     "promise",
     "standard",
     "react"
-  ]
+  ],
+  "rules": {
+    "yoda": [2, "never", { "exceptRange": true }]
+  }
```
devvmh: I really like using `if (1 <= variable && variable <= 5)` as a way to check if a variable is in a given range. This is a small change to standard's eslint rules to allow this.

PR #718, opened 2016-10-04
fix api documentation

devvmh:


Connoropolous:
let's put email onto current user... possibly some other things too, let me think...

PR #719, opened 2016-10-04
add starred to maps API

devvmh:
plus a few other bonus commits

### DIFF
--- a/app/controllers/api/v2/stars_controller.rb
+++ b/app/controllers/api/v2/stars_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+module Api
+  module V2
+    class StarsController < RestfulController
+      skip_before_action :load_resource
+
+      def create
+        @map = Map.find(params[:id])
+        @star = Star.new(user: current_user, map: @map)
+        authorize @map, :star?
+        create_action
+
+        if @star.errors.empty?
+          render json: @map, scope: default_scope, serializer: MapSerializer, root: serializer_root
+        else
+          respond_with_errors
+        end
+      end
+
+      def destroy
+        @map = Map.find(params[:id])
+        authorize @map, :unstar?
+        @star = @map.stars.find_by(user: current_user)
+        @star.destroy if @star.present?
+        head :no_content
+      end
+    end
+  end
+end
--- a/app/models/star.rb
+++ b/app/models/star.rb
@@ -2,4 +2,5 @@
 class Star < ActiveRecord::Base
   belongs_to :user
   belongs_to :map
+  validates :map, uniqueness: { scope: :user, message: 'You have already starred this map' }
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -65,7 +65,10 @@ Metamaps::Application.routes.draw do
     namespace :v2, path: '/v2' do
       resources :metacodes, only: [:index, :show]
       resources :mappings, only: [:index, :create, :show, :update, :destroy]
-      resources :maps, only: [:index, :create, :show, :update, :destroy]
+      resources :maps, only: [:index, :create, :show, :update, :destroy] do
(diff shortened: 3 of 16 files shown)


app/controllers/api/v2/stars_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      skip_before_action :load_resource
+
+      def create
+        @map = Map.find(params[:id])
+        @star = Star.new(user: current_user, map: @map)
+        authorize @map, :star?
+        create_action
+
+        if @star.errors.empty?
+          render json: @map, scope: default_scope, serializer: MapSerializer, root: serializer_root
```
devvmh: In a theoretical world, this should actually return an instance of a star resource... but that's kinda dumb, so I decided to return the new map instead

config/routes.rb
```diff
@@ -65,7 +65,10 @@
     namespace :v2, path: '/v2' do
       resources :metacodes, only: [:index, :show]
       resources :mappings, only: [:index, :create, :show, :update, :destroy]
-      resources :maps, only: [:index, :create, :show, :update, :destroy]
+      resources :maps, only: [:index, :create, :show, :update, :destroy] do
+        post :stars, to: 'stars#create', on: :member
+        delete :stars, to: 'stars#destroy', on: :member
```
devvmh: If you use resources :stars, only: [:create, :destroy], then it'll make you use the star id in the URL for deletion, which is dumb

doc/api/api.raml
```diff
@@ -1,7 +1,7 @@
 #%RAML 1.0
 ---
 title: Metamaps
-version: v2
+version: v2.0
```
devvmh: add a minor version so once we publish to metamaps.cc, we can bump this to 2.1 if we add new (backwards-compatible) functionality

app/controllers/api/v2/stars_controller.rb
Connoropolous: Hmm, my aim has been to ensure that there is only ever one Star connecting a user and a map... in theory you could keep hitting this endpoint and create multiple right? I'm not sure what a good way to prevent that would be

config/routes.rb
Connoropolous: Agreed

doc/api/api.raml
Connoropolous: Good thinking

app/models/star.rb
```diff
@@ -2,4 +2,5 @@
 class Star < ActiveRecord::Base
   belongs_to :user
   belongs_to :map
+  validates :map, uniqueness: { scope: :user, message: 'You have already starred this map' }
```
Connoropolous: Oh, this takes care of my other comment about uniqueness. Got it :)

doc/api/examples/map.json
```diff
@@ -5,6 +5,7 @@
         "desc": "Example map for the API",
         "permission": "commons",
         "screenshot": "https://s3.amazonaws.com/metamaps-assets/site/missing-map.png",
+        "starred": false,
```
Connoropolous: I guess this will be included even if you're signed out? Is there a way to make it behave like 'is_admin' on current_user?

devvmh: Ok i can do that

I just dont like changing the schema too much you know? I suppose it might happen more often with these complex fields... 

I was thinking though that as a programmer, its more helpful for this to exist in the response than not exist.

I suppose i could make.its value null. But i actually think i prefer this way

devvmh: And by changing the schema,  i mean making the json different in different situations. I'd like the shape of the responses to be as similar as possible

Connoropolous: yeye. Ok I am happy to go that direction, with the background you've provided :) leave as is

PR #724, opened 2016-10-05
Made it so that the map pans when you drag a selection of topics near the edge of the window

Bortseb:


devvmh:
Looks great! Should we test or is this ready to merge?

Bortseb:
I've tested a bunch... but it could still use a look over.. maybe even moreso the code changes I've made... there may be recommendations for how to more efficiently achieve the same result.

frontend/src/Metamaps/JIT.js
```diff
@@ -765,71 +769,109 @@ const JIT = {
...
+            Metamaps.Visualize.mGraph.canvas.translate(SHIFT,0);
+            self.virtualPointer = {x:self.virtualPointer.x - SHIFT,y:self.virtualPointer.y};
+            self.updateTopicPositions(node,self.virtualPointer);
+            Visualize.mGraph.plot();
+          } , PERIOD);
+        }
+        if(width - xPix < EDGE_THICKNESS && !self.hasOwnProperty('dragRightEdge') && self.dragTolerance){
+          Metamaps.Visualize.mGraph.canvas.translate(-SHIFT,0);
+          self.virtualPointer = {x:pos.x + SHIFT,y:pos.y};
+          self.updateTopicPositions(node, self.virtualPointer);
```
Connoropolous: nice work @Bortseb!
I tried out the implementation on your C9 instance and it works great, for the most part! I notice it can be a bit 'glitchy' visually, and I think there might be some kind of change that we can do to the above three lines to resolve the glitchy-ness. 

right now the approach is: 
translate the canvas
move the topic by the exact same amount as the canvas was translated. this is good in theory, but as we can see, glitchy in practice, possibly because of the setTimeouts. 

our other option is: 
translate the canvas
use `util.pixelsToCoords` to calculate what the position of the cursor over the canvas is equal to presently
use updateTopicPositions to update the node to the coordinate returned by the last function call

let's give that a try? 
I also think we can increase the dimension of the dragzones at the edges of the screen a little bit :)

devvmh:
I just tried this out, and this is great! Even if we can't iron out all the glitches, I think it should definitely get in by Oct 31

devvmh:
i added a merge commit to remove the conflicts with develop. We can revert that commit if I broke anything.

Bortseb:
@Connoropolous have a look at this PR now... I think it is significantly improved

Connoropolous:
@bortseb can't review it right now because of all the merge conflicts, it doesn't seem up to date with 'develop' :s take a look at the 'changes' tab

Bortseb:
Whoa! What the!

PR #731, opened 2016-10-06
dont do async: false

### LINKED ISSUE #684: pulling in topics from autocomplete uses deprecated synchronous ajax call
screen shot 2016-09-28 at 8 48 46 pm

make this happen async instead

Connoropolous:
remove fetchContained wasn't doing anything anyways

fixes #684

### DIFF
--- a/frontend/src/Metamaps/AutoLayout.js
+++ b/frontend/src/Metamaps/AutoLayout.js
@@ -49,7 +49,7 @@ const AutoLayout = {
       }
     }
 
-    if (opts.map && self.coordsTaken(nextX, nextY, opts.map)) {
+    if (opts.mappings && self.coordsTaken(nextX, nextY, opts.mappings)) {
       // check if the coordinate is already taken on the current map
       return self.getNextCoord(opts)
     } else {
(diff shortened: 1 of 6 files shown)


frontend/src/Metamaps/AutoLayout.js
```diff
@@ -59,8 +59,7 @@ const AutoLayout = {
       }
     }
   },
-  coordsTaken: function (x, y, map) {
-    const mappings = map.getMappings()
+  coordsTaken: function (x, y, mappings) {
```
devvmh: nice

frontend/src/Metamaps/Topic.js
```diff
@@ -35,30 +35,14 @@ const Topic = {
...
-      }
-    }
+      $.ajax({
+        url: '/topics/' + id + '.json',
+        success: function (data) {
+          Metamaps.Topics.add(data)
+          callback(Metamaps.Topics.get(id))
+        }
+      })
+    } else callback(Metamaps.Topics.get(id))
```
devvmh: what if callback is null?

Connoropolous: should I catch that case then?

devvmh: i guess we could do

`const identityFunction = x => x`
`callback = identifyFunction`

devvmh: (as a default argument)

PR #733, opened 2016-10-06
dont pan while using arrow keys during creation fixes #721

### LINKED ISSUE #721: When using arrow keys in topic autocomplete, entire map shifts up and down.

Connoropolous:


devvmh:
Hmmm. It might be better to figure out a good way to tell when key combos are being run on the canvas/body instead of on child elements. E.g. what about scrolling in a lightbox? That might also pan the map

Of course this us great for now!

Connoropolous:
Yes, I agree, there are too many cross-overs, it will get unmanageable very quickly.

Connoropolous:
ack. @devvmh .... there's a bigger issue with the using arrow keys to pan... any time you are in a text field of any kind, and use the arrow keys: search box, fork map fields, message field, etc. 
it is always panning the map. we have to disable arrow keys to pan, or find a way to fix these conflicts

PR #734, opened 2016-10-06
dont show private maps in global collection

Connoropolous:
@devvmh should I be accessing Perm::Ission to get the 'Private' string?

### DIFF
--- a/app/controllers/explore_controller.rb
+++ b/app/controllers/explore_controller.rb
@@ -9,7 +9,7 @@ class ExploreController < ApplicationController
 
   # GET /explore/active
   def active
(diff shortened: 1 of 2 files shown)


devvmh:
Haha, debatable. I mostly made that service to factor out the functions, and put the strings in there. We didn't even have policies then. It's hard to say. I wouldn't worry about it unless we add a fourth permission category!

devvmh:
Does it still work if Private is capitalized?

Connoropolous:
ah... I think it does, but it should be switched to lower case. Doing that now and merging

PR #738, opened 2016-10-07
Makes it so that resizing the browser window doesn't change the user's location on the map

Bortseb:


frontend/src/Metamaps/Listeners.js
```diff
@@ -122,7 +123,24 @@ const Listeners = {
...
+          centreCoords = Util.pixelsToCoords({x:centrePixX ,y:centrePixY});
+        
+        Visualize.mGraph.canvas.resize($(window).width(), $(window).height())
+  
+        canvas.scale(scaleX,scaleY);
+        var newCentrePixX = Visualize.mGraph.canvas.canvases[0].size.width / 2,
+            newCentrePixY = Visualize.mGraph.canvas.canvases[0].size.height / 2,
+            newCentreCoords = Util.pixelsToCoords({x:newCentrePixX ,y:newCentrePixY});
+        
+        canvas.translate(newCentreCoords.x - centreCoords.x, newCentreCoords.y - centreCoords.y);
```
Connoropolous: @Bortseb can you describe all of this a little bit? what is the behaviour that all this enables? in particular in terms of centering?

Bortseb: ok @Connoropolous I commented the code. Let me know if this is what you wanted.

Connoropolous: yeah, I wasn't thinking in the code, but in the code is fine and good too!

frontend/src/Metamaps/Listeners.js
```diff
@@ -122,7 +123,27 @@ const Listeners = {
     })
 
     $(window).resize(function () {
-      if (Visualize && Visualize.mGraph) Visualize.mGraph.canvas.resize($(window).width(), $(window).height())
+      if (Visualize && Visualize.mGraph){
+        //Find the current canvas scale and map-coordinate at the centre of the user's screen
+        var canvas = Visualize.mGraph.canvas,
```
Connoropolous: All further references to `Visualize.mGraph.canvas` after this variable declaration should just be simplified to `canvas`

Bortseb: Ok done

Bortseb:
Sure! Share away...  Apparently I was eager to push this one out there..
And missed some pretty basic stuff, so why not learn at every stage so it
gets drilled in :)

Connoropolous:
Cool! I will never generally speaking not hesitate to share then :) As much as the development of these features/bug fixes is great, building skilled development capacity within our network is priceless :p 

So the thing to consider here is: in the $window.resize event, how easy in the future would it be to add new code? It was easy for you to add the code that you added, but what if we needed new functions that also got called when window resize events were firing. That function could start to get pretty messy. So we want to make it easy for that next person even as we're adding the code we're adding. 

that just means looking for a good place to outsource that functionality to... then calling those functions from within the resize event.

devvmh:
I pushed a commit with code style changes; if I broke anything we can revert my commit.

devvmh:
Right, can you make this code into a function
```
Visualize.safeResize = function(width, height) {
  // your code here
}
```
and then call the function in Listeners.js?

otheriwse LGTM

Bortseb:
Just a note to self... this is what I was originally intending:

> So, maybe I create a function called logCanvasAttributes that returns an JS object with all the stuff I need... then I can just call this function before and after the  resize..

Bortseb:
Hows this look now @Connoropolous ?

devvmh:
This looks great. Lets merge it!

Connoropolous:
nay! I'd like to go further with it still though actually, so that the `resize` function stays clean, which we're still not really achieving here. So move more of the code into the util function maybe, and then just call it from the resize function and pass in the `canvas`?

```
$(window).resize(function () {
    Util.resizeCanvas(Visualize.mGraph.canvas)
    if (Active.Map && Realtime.inConversation) Realtime.positionVideos()
    Mobile.resizeTitle()
})		     
```

Connoropolous:
because it's a reasonably often  occurence to have something else come up that gets triggered on window resize

devvmh:
Sounds good. I still say merge now and refactor later though

Maybe Map or Visualize cpuld host one of the functions

Connoropolous:
to you both, to be clear, I'd be totally happy and fine for it to go in as is :)

it is especially in regards to a short check in with Rob where we discussed not rushing, in order to share more knowledge and be talking about what the best approaches to things are :)
happy however it goes here

Bortseb:
pretty neat to watch live chat happen here :p

Connoropolous:
super

PR #739, opened 2016-10-07
[onhold] copy/paste topics and synapses

devvmh:
the issue as framed from the users perspective: https://trello.com/c/r5FVpkZj/23-cut-copy-paste-of-topics-and-synapses-between-maps

This mostly works. It's still a bit awkward, because the x & y positions are included in the paste. This is annoying if you paste in the same map, since it pastes on top of existing topics, and when you try to move the selection, it moves the old, unselected topics

But if you're willing to work around the UX problems it's working well.

Should this be included in 3.0?

### DIFF
--- a/app/controllers/maps_controller.rb
+++ b/app/controllers/maps_controller.rb
@@ -116,8 +116,12 @@ class MapsController < ApplicationController
 
   # GET maps/:id/export
   def export
-    exporter = MapExportService.new(current_user, @map, base_url: request.base_url)
-
+    topic_ids = params[:topic_ids].split(',').map(&:to_i)
+    synapse_ids = params[:synapse_ids].split(',').map(&:to_i)
+    exporter = MapExportService.new(current_user, @map,
+                                    topic_ids: topic_ids,
+                                    synapse_ids: synapse_ids,
+                                    base_url: request.base_url)
     respond_to do |format|
       format.json { render json: exporter.json }
       format.csv { send_data exporter.csv }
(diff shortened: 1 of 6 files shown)


devvmh:
This is too gross for 3.0, so I'll bump to 3.5

devvmh:
I'm considering the best next step for this PR. The big issue is 1) not swarming the database with thousands of copied topics and 2) currently if you copy-paste into the same map, the topics are on top of each other. I was considering adding metadata to the JSON and/or popping up dialogs asking the user to choose how best to import the data, but as I wrote it I started thinking that maybe the best next step is this:

To begin, use our current JSON export but add an (optional) options hash in the JSON export/import. for instance, you might have

```json
{
  topics: [{ name: 'Topic1', x: 0, y: 5 }],
  synapses: [],
  options: {
    originalMap: {
      url: 'https://metamaps.cc/maps/5',
      xOffset: -50,
      yOffset: 0,
      scale: 1
    }
  }
}
```

Then we could match when importing if it was the same map (by base url and map id). If it was the same map, we would create new topics and *ignore* the xOffset and yOffset params, pasting the topics in the same arrangement but centred around a new map position (e.g. you copy-paste a topic, pan the map to the left, and then paste, and the topics will appear to the left of the originals the same amount that you panned)

If importing into a different map on the same server, we could "pull in" the topics instead. Probably we would want to paste the topics relative to the original offset though.

@harlantwood @Bortseb @Connoropolous please weigh in with thoughts 😁

PR #747, opened 2016-10-08
redirect to root_path if you get a 403

devvmh:
Already hotfixed on production


Connoropolous:
oof. good one, thankyou! +R

PR #756, opened 2016-10-12
suggesting api doc updates

devvmh:


doc/api/pages/getting-started.md
```diff
@@ -25,13 +25,15 @@ fetch('/api/v2/tokens', {
...
+```javascript
 fetch('/api/v2/tokens', {
   method: 'POST',
   credentials: 'same-origin'
 }).then(response => {
   return response.json()
-}).then(console.log).catch(console.error)
+}).then(payload => {
+  console.log(payload)
+}).catch(console.error)
```
devvmh: I like this way because then I can refer to `payload` on the next line

PR #762, opened 2016-10-14
access requests

Connoropolous:
this PR also enables sending someone a link like
https://metamaps.cc/users/sign_up?code=iuytrewq&redirect_to=/maps/5
and allowing another person to sign up with their code and go directly to their map

- [x] also a push email when the map creator approves the access request. 
- [x] an indicator on 'view-only' maps (public where you can't edit) to request edit access

if the map is private and you try and hit /maps/42 and you are signed out, you redirect to sign in
if the map is private and you try and hit /maps/42 and you're signed in, you redirect to /request_access
once you request access, it redirects you back to the explore maps pages
screen shot 2016-10-13 at 10 42 18 pm

if the map is public and you're signed in and you visit it:
screen shot 2016-10-16 at 7 21 36 pm

if you click request access (and if you refresh the page you still see 'request pending'
screen shot 2016-10-16 at 7 21 45 pm

if the map is public or commons and you're signed out and you visit it:
screen shot 2016-10-16 at 7 23 00 pm

the map creator receives an email
screen shot 2016-10-13 at 10 39 44 pm

what happens when you click "grant"
screen shot 2016-10-13 at 10 43 12 pm

the person who made the request receives an email
screen shot 2016-10-16 at 8 14 37 pm


app/controllers/maps_controller.rb
```diff
@@ -93,7 +112,7 @@ def access
 
     respond_to do |format|
       format.json do
-        render json: { message: 'Successfully altered edit permissions' }
+        head :ok
```
devvmh: why this change? Sending json seems better to me

app/controllers/maps_controller.rb
```diff
@@ -133,6 +152,48 @@ def events
     end
   end
 
+  # GET maps/:id/approve_access/:request_id
+  def approve_access
+    request = AccessRequest.find_by_id(params[:request_id])
```
devvmh: this can just be AccessRequest.find(params[:request_id]), the find_by helpers are deprecated

You can also do AccessRequest.find_by(id: ...) or AccessRequest.find_by(name: ...), etc, which is why they took the named helpers out

app/views/maps/request_access.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    var that = $(this)
+    that.off('click')
+    that.text('requesting...')
+    $.ajax({
+      url: '/maps/<%= params[:id] %>/access_request',
+      type: 'POST',
+      contentType: 'application/json',
+      statusCode: {
+        200: function () { that.text('Request Sent!'); setTimeout(function () {window.location.href = '/'}, 2000) },
+        500: function () { that.text('An error occurred') }
```
devvmh: I could be wrong but I think `head :bad_request` is a 400 code

app/controllers/maps_controller.rb
Connoropolous: It seems random, and not really necessary to send a string of text. We don't use that string anywhere in the UI.

app/controllers/maps_controller.rb
Connoropolous: Hmm, kk. But, the reason that I went with this was specifically because AccessRequest.find throws an error when it doesn't find one, which i don't want. Hmm or maybe I do. I was thinking of just redirecting them and displaying the message as a nicer option, but maybe it's unnecessary, given that the times when someone would hit this URL and not find one is incrediblly rare, they should only ever hit this url by clicking through "Accept" or "Deny" links out of an email...

app/views/maps/request_access.html.erb
Connoropolous: Right you are

app/controllers/maps_controller.rb
```diff
@@ -133,6 +152,48 @@ def events
...
+    request.deny() if request
+    respond_to do |format|
+      format.json do
+        head :bad_request unless request
+        head :ok
+      end
+    end
+  end
+
```
Connoropolous: Yeah, that seems pretty solid. I will do that

Connoropolous:
What still needs doing down the road: 
- [ ] a push notification to the browser to the map creator, and a permanent notification
- [ ] a push notification to the person who requested access, and a permanent notification

Connoropolous:
@devvmh made all your style changes :)

PR #767, opened 2016-10-17
schema update

devvmh:
@connoropolous what do you think of this change? It happened automatically when I ran db migrate....

Should I consult production? merge this? close this? Maybe matching things to the production schema is best, I'm not sure.

I guess we could upgrade postgres on the production server at some point too.


Connoropolous:
aha, yes, let's merge this. i realized after i had already ran my own migration that I should set the defaults, so that's why schema.rb didn't get updated with those values.

devvmh:
ok cool. just a note for later: postgresql is v9.3 on the server

PR #768, opened 2016-10-17
add scripts from default rails install

devvmh:
I ran `rspec` just now and my computer said:

```
Migrations are pending. To resolve this issue, run:

  bin/rails db:migrate RAILS_ENV=test
```

but I couldn't, because bin/rails didn't exist. So I ran `rails new` in another folder and copied these three scripts over.


Connoropolous:
I don't exactly understand what these are and what they're for, but it seems like it would make sense that they go in?

devvmh:
Yeah, I never use them. I think it's supposed to be a guarantee that where `rake db:migrate` might fail on some quirky setups, `bin/rake db:migrate` should be more reliable

PR #770, opened 2016-10-17
remove check-canvas-support.js

devvmh:
This function was added in a commit from 2012. It's hard to say, but I think it
was stolen from the JIT code. It seems like older iPhones used an older API for
touch events. I have two reasons for thinking it's OK to remove this:

1. It looks like Safari on mobile just works (TM) with these events now, along with other browsers (although this isn't totally clear to me). See https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#Browser_compatibility
2. The four variables declared in this function aren't referenced anywhere else in our code base.

It's possible I'm wrong; if we can find one, we should try it out on an older model iPhone and maybe a few older androids. If we can try on an iPad too that would be nice


Connoropolous:
no, you're correct, we no longer need it :)

devvmh:
yolo

PR #771, opened 2016-10-17
multiple policy issues

Connoropolous:
this should actually prbly be hotfixed into metamaps.cc as well

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -65,6 +65,10 @@ class User < ApplicationRecord
     json
   end
   
+  def all_accessible_maps
+    maps + shared_maps
+  end
+
   def recentMetacodes
     array = []
     self.topics.sort{|a,b| b.created_at <=> a.created_at }.each do |t|
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -66,6 +66,7 @@ class User < ApplicationRecord
   end
   
   def all_accessible_maps
+    #TODO: is there a way to keep this an ActiveRecord relation?
     maps + shared_maps
   end
 
--- a/app/policies/mapping_policy.rb
+++ b/app/policies/mapping_policy.rb
@@ -10,6 +10,8 @@ class MappingPolicy < ApplicationPolicy
       permission = 'maps.permission IN (?)'
       return scope.joins(:map).where(permission, visible) unless user
 
+      # if this is getting changed, the policy_scope for messages should also be changed
+      # as it is based entirely on the map to which it belongs
       scope.joins(:map).where(permission, visible)
            .or(scope.joins(:map).where('maps.id IN (?)', user.shared_maps.map(&:id)))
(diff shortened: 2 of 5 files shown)


app/models/user.rb
```diff
@@ -65,6 +65,10 @@ def as_json_for_autocomplete
     json
   end
   
+  def all_accessible_maps
+    maps + shared_maps
```
devvmh: `maps.or(shared_maps)` is better because it stays as an activerecord relation

app/policies/mapping_policy.rb
```diff
@@ -8,11 +8,11 @@ def resolve
...
-        scope.joins(:map).where(permission, visible).or(scope.joins(:map).where(user_id: user.id))
-      else
-        scope.joins(:map).where(permission, visible)
-      end
+      return scope.joins(:map).where(permission, visible) unless user
+
+      scope.joins(:map).where(permission, visible)
+           .or(scope.joins(:map).where('maps.id IN (?)', user.shared_maps.map(&:id)))
+           .or(scope.joins(:map).where('maps.user_id = ?', user.id))
     end
```
Connoropolous: So the policy_scope for `mappings` and `messages` is the same... it is entirely based off the map to which it belongs

app/policies/synapse_policy.rb
```diff
@@ -7,7 +7,7 @@ def resolve
       return scope.where(permission: visible) unless user
 
       scope.where(permission: visible)
-           .or(scope.where(defer_to_map_id: user.shared_maps.map(&:id)))
+           .or(scope.where(defer_to_map_id: user.all_accessible_maps.map(&:id)))
```
devvmh: brilliant. can we add a comment explaining this mind bending permission logic?

app/policies/synapse_policy.rb
```diff
@@ -7,7 +7,7 @@ def resolve
       return scope.where(permission: visible) unless user
```
Connoropolous: in the topics and synapses policy_scopes, these changes are to fix the circumstances where:
you want to see a topic or synapse which
- is not `commons`
- you did not create
- has a `defer_to_map_id` value which references a map that is not "shared with you" but that you created

app/policies/synapse_policy.rb
devvmh: maybe something like "map is visible if 1) it's public/commons, 2) it's defer_to_map_id isn't nil and points to a map visible to the user, or 3) the user owns the map"

maybe we should even make that note about not being nil explicit in the code

    .or(scope.where.not(defer_to_map_id: nil).where(defer_to_map_id: user.all_accessible_maps.map(&:id))

app/models/user.rb
Connoropolous: this somehow combines the two? even though its an 'or'?

PR #772, opened 2016-10-17
access request tests

devvmh:


spec/mailers/map_mailer_spec.rb
```diff
@@ -1,10 +1,24 @@
...
+    let(:request) { create(:access_request) }
+    let(:map) { create(:map) }
+    let(:mail) { described_class.access_request_email(request, map) }
+
+    it { expect(mail.from).to eq ['team@metamaps.cc'] }
+    it { expect(mail.to).to eq [map.user.email] }
+    it { expect(mail.subject).to match map.name }
+    it { expect(mail.body.encoded).to match map.name }
+    it { expect(mail.body.encoded).to match request.user.name }
+    it { expect(mail.body.encoded).to match 'Approve Request' }
```
Connoropolous: the phrasing is actually just 'Grant' and 'Deny'

Would that make this fail?

devvmh: It's in there somewhere! I'm not sure why. I'd have to go back to the templates. But they do pass right now

Connoropolous: oh, it's in the plain text version

Issue #570, opened 2016-07-29
make video chat stable


devvmh:

Connoropolous:
Summary of Findings from todays testing with Ben: 
When you leave a map to Explore, then join a map, Video doesnt work at all
When you start on a map, and refresh, video works, but unreliably

https://docs.google.com/document/d/1SMPORTdv_3dh9CzyF0NCt3oKP7-jczNxdpsuaTHqYQc/edit

Connoropolous:
Could it be the lack of a TURN server? or the particular STUN server we're using? 
https://github.com/metamaps/metamaps/blob/develop/realtime/realtime-server.js#L4

Other possible stun servers, and even some turn servers
https://gist.github.com/yetithefoot/7592580

Connoropolous:
@devvmh stabilizing junto is going to be something I can't do alone :s 
it would be so much better if we could take a stab at it together, see http://piratefsh.github.io/projects/2015/08/27/webrtc-stun-turn-servers.html

there needs to be more infrastructure for it to become reliable

Connoropolous:
an alternative is coughing up some more money and relying on http://xirsys.com/ but I don't feel great about that route...

Connoropolous:
fixed by
#779 and #780 

will reopen if need be, and doesn't prove stable

PR #779, opened 2016-10-18
holy insanity realtime refactor

Connoropolous:
just one or two little tweaks left to make

requires running the realtime-server.js file with babel-node
`babel-node realtime/realtime-server.js


Connoropolous:
remarkably, ironically, and extremely satisfactorily, I believe this also solves the primary huge bug we've been having

(it's about 2 lines of code that makes the difference)

Connoropolous:
#570 and #712

PR #781, opened 2016-10-19
clean up and fix devise code

devvmh:
Fixes a sign in bug and also cleans up the routes to clearly point everything to our controllers rather than the base devise controllers


devvmh:
It's possible this will cause bugs, but that's why we have the week of bug fixes next week 😁

PR #791, opened 2016-10-21
[WIP] keep client fresh with junto info globally

Connoropolous:


.babelrc
```diff
@@ -5,6 +5,8 @@
   ],
   "plugins": [
     "lodash",
-    "transform-class-properties"
+    "transform-class-properties",
+    "transform-es2015-destructuring",
```
devvmh: this should be included in babel-preset-es2015

.babelrc
```diff
@@ -5,6 +5,8 @@
...
+    "transform-object-rest-spread"
```
devvmh: I do think we should remove this one. Especially since it's not even in the es2016 spec, it's still in draft stage 3, so it's possible it'll change

devvmh:
ok pushed some commits.

using redux is so great, it just gives us so much more certainty about the state of the video. if only we could do the same with all those listeners!

Connoropolous:
Awesome! thanks heaps for that

PR #796, opened 2016-10-22
fixed the oauth redirect bug and a bit more

### LINKED ISSUE #502: ensure that oauth flow keeps your sign in process, even if you weren't signed in
known behaviour: almost forgot about this

if you're trying to sign in to another platform with metamaps, or get a token, the flow only works right now if you're already signed in. 

If you're not signed in, it redirects you to do that, but doesn't redirect you back into your oauth flow

Connoropolous:
hid "register an app" from non-admins
added "apps" to the account menu, as a teaser

@devvmh the only thing I'm not 100% certain of is possible adverse side effects from adding a bit of duped code into 
app/controllers/users/sessions_controller.rb

screen shot 2016-10-22 at 1 45 52 am

fixes #502

### DIFF
--- a/app/assets/stylesheets/application.scss.erb
+++ b/app/assets/stylesheets/application.scss.erb
@@ -827,6 +827,9 @@ label {
 .accountAdmin .accountIcon {
     background-position: 0 -32px;
 }
+.accountApps .accountIcon {
+    background-position: 0 -32px;
+}
 .accountInvite .accountIcon {
     background-position: 0 -64px;
 }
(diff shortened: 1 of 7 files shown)


Connoropolous:
the root of the problem was overly simplistic functions in config/initializers/doorkeeper.rb

Connoropolous:
@devvmh if you read this one and dig it, just merge it :)

devvmh:
LGTM

PR #839, opened 2016-10-25
make synapse permissions depend on topic1 and topic2

devvmh:


spec/controllers/synapses_controller_spec.rb
```diff
@@ -35,13 +35,30 @@
       end
     end
 
+    context 'with private topics' do
+      it 'redirects to /' do
```
Connoropolous: Why redirect to / on a json post request?

app/controllers/synapses_controller.rb
```diff
@@ -22,10 +22,15 @@ def create
     @synapse = Synapse.new(synapse_params)
     @synapse.desc = '' if @synapse.desc.nil?
     @synapse.desc.strip! # no trailing/leading whitespace
-    authorize @synapse
+
+    success = nil
+    ActiveRecord::Base.transaction do
```
Connoropolous: What does this do?

app/policies/synapse_policy.rb
```diff
@@ -16,16 +16,11 @@ def index?
   end
 
   def create?
```
Connoropolous: This all looks good

Connoropolous: There is always there fact that the permission of the topics which a synapse connects were accessible at the time it was created and then get changed, so we still have to bear those scenarios

Connoropolous: beware*

app/controllers/synapses_controller.rb
devvmh: Here's the issue: I was getting unathorized when I gave invalid params. I'd rather get unprocessable entity when i give invalid params. So I have to run @synapse.save, and THEN authorize. If it doesn't authorize, I want it to roll back. So the ActiveRecord transaction will run SQL commands + rails commands, and if it errors out it will roll back the transaction and never write it to disk

spec/controllers/synapses_controller_spec.rb
devvmh: Because of how we handle Pundit::NotAuthorized in application controller - we don't return 403, we do conditional redirects. I was surprised too, but it is how it works so the test is at least documenting that!

devvmh:
Cool I'll leave this open for now.

We could also drop the update to `create?` in synapse policy to make it less complicated.

app/controllers/synapses_controller.rb
```diff
@@ -22,10 +22,15 @@ def create
...
+      success = @synapse.save
+      success ? authorize(@synapse) : skip_authorization
```
devvmh: Now that I read this line on Github I should probably make it clearer. Originally I had it as

```
authorize(@synapse) if success
skip_authorization if !success
```

spec/controllers/synapses_controller_spec.rb
Connoropolous: Oh, hmm. That's not great. Get requests only should do that, I am thinking

Connoropolous: I updated that not long ago and thought I made an improvement, but only for some cases :p

app/controllers/synapses_controller.rb
Connoropolous: Got it, cool, I'm glad it's documented in this PR :)

spec/controllers/synapses_controller_spec.rb
devvmh: I think it's OK for now, no need to rush. maybe we can make an issue

devvmh: as we fix these things too, we can "document" the behaviour by writing tests

Connoropolous: Yes, true. And actually, I am thinking that in cases like synapse creation, itd be good to start using our own API :) then just the main app handle page loads and special functions

PR #840, opened 2016-10-25
policy_scope on has_many relationships in serializers

### LINKED ISSUE #818: api embed doesn't scope relationships
E.g. http://metamaps.herokuapp.com/api/v2/maps/132?embed=topics returns private topics

devvmh:
fixes #818

### DIFF
--- a/app/controllers/api/v2/restful_controller.rb
+++ b/app/controllers/api/v2/restful_controller.rb
@@ -45,7 +45,7 @@ module Api
       end
 
       def current_user
-        super || token_user || doorkeeper_user || nil
+        token_user || super || doorkeeper_user || nil
       end
 
       def load_resource
--- a/app/serializers/api/v2/application_serializer.rb
+++ b/app/serializers/api/v2/application_serializer.rb
@@ -32,10 +32,10 @@ module Api
           if attr.to_s.pluralize == attr.to_s
             attribute("#{attr.to_s.singularize}_ids".to_sym,
                       opts.merge(unless: -> { embeds.include?(key) })) do
-              Pundit.policy_scope(scope[:current_user], object.send(:topics)).map(&:id)
+              Pundit.policy_scope(scope[:current_user], object.send(attr)).map(&:id)
             end
             has_many(attr, opts.merge(if: -> { embeds.include?(key) })) do
-              Pundit.policy_scope(scope[:current_user], object.send(:topics))
+              Pundit.policy_scope(scope[:current_user], object.send(attr))
             end
           else
             id_opts = opts.merge(key: "#{key}_id")
--- a/app/serializers/api/v2/application_serializer.rb
+++ b/app/serializers/api/v2/application_serializer.rb
@@ -32,10 +32,10 @@ module Api
           if attr.to_s.pluralize == attr.to_s
             attribute("#{attr.to_s.singularize}_ids".to_sym,
                       opts.merge(unless: -> { embeds.include?(key) })) do
-              Pundit.policy_scope(scope[:current_user], object.send(attr)).map(&:id)
+              Pundit.policy_scope(scope[:current_user], object.send(attr))&.map(&:id) || []
             end
             has_many(attr, opts.merge(if: -> { embeds.include?(key) })) do
-              Pundit.policy_scope(scope[:current_user], object.send(attr))
+              Pundit.policy_scope(scope[:current_user], object.send(attr)) || []
             end
           else
             id_opts = opts.merge(key: "#{key}_id")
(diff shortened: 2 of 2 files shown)


app/controllers/api/v2/restful_controller.rb
```diff
@@ -45,7 +45,7 @@ def accessible_records
       end
 
       def current_user
-        super || token_user || doorkeeper_user || nil
+        token_user || super || doorkeeper_user || nil
```
devvmh: This way if you're logged in but have an access token, you can override the current_user when using the browser to access the api. mostly for me but i can see it being useful for other developers

app/serializers/api/v2/application_serializer.rb
```diff
@@ -32,9 +32,11 @@ def self.embed_dat
...
             attribute("#{attr.to_s.singularize}_ids".to_sym,
                       opts.merge(unless: -> { embeds.include?(key) })) do
-              object.send(attr).map(&:id)
+              Pundit.policy_scope(scope[:current_user], object.send(:topics)).map(&:id)
+            end
+            has_many(attr, opts.merge(if: -> { embeds.include?(key) })) do
+              Pundit.policy_scope(scope[:current_user], object.send(:topics))
             end
-            has_many(attr, opts.merge(if: -> { embeds.include?(key) }))
           else
```
devvmh: note has_one is unchanged

app/controllers/api/v2/restful_controller.rb
Connoropolous: Makes sense

app/serializers/api/v2/application_serializer.rb
Connoropolous: This doesn't make sense to me, why is topics being hard coded?

devvmh:
Here's the comment from the code itself:

```
      # self.embeddable might look like this:
      #   creator: { attr: :first_creator, serializer: UserSerializer }
      #   contributors: { serializer: UserSerializer}
      # This method will remove the :attr key if the underlying attribute name
      # is different than the name provided in the final json output. All other keys
      # in the hash will be passed to the ActiveModel::Serializer `attribute` method
      # directly (e.g. serializer in the examples will be passed).
      #
      # This setup means if you passed this self.embeddable config and sent no
      # ?embed= query param with your API request, you would get the regular attributes
      # plus creator_id and contributor_ids. If you passed ?embed=creator,contributors
      # then instead of an id and an array of ids, you would get a serialized user
      # (the first_creator) and an array of serialized users (the contributors).
```

So walking through the code a bit, here's what's going on:

```
      def self.embed_dat
        embeddable.each_pair do |key, opts|
          attr = opts.delete(:attr) || key
          if attr.to_s.pluralize == attr.to_s
            attribute("#{attr.to_s.singularize}_ids".to_sym,
                      opts.merge(unless: -> { embeds.include?(key) })) do
              Pundit.policy_scope(scope[:current_user], object.send(attr))&.map(&:id) || []
            end
            has_many(attr, opts.merge(if: -> { embeds.include?(key) })) do
              Pundit.policy_scope(scope[:current_user], object.send(attr)) || []
            end
          else
            id_opts = opts.merge(key: "#{key}_id")
            attribute("#{attr}_id".to_sym,
                      id_opts.merge(unless: -> { embeds.include?(key) }))
            attribute(key, opts.merge(if: -> { embeds.include?(key) }))
          end
        end
      end
```

`embeddable` is an object with keys. The keys are what is used to control the `?embed=` query param; for the embed to work, the text  has to match one of the keys.

Now i can iterate over `embeddable`, which is a mapping of those keys to extra options. If you look through the serializers, generally it just looks like `{ user: {}, map: {} }`. That is, I'm not user any extra options (`opts` in the code). But I am still passing that key.

Now we move into the each_pair block. You can pass an attribute name (synapses used to have `{ topic1: { attr: :node1 } }`, etc, to translate the db column name to the api name. So `key` is the "api name", and `attr` is the "database name".

Now that we have attr and key, we need to figure out if this is has_many or has_one. The if statement does that by basically checking if the attr is pluralized. Top condition is has_many, bottom condition is has_one.

Now we get two calls. Because `self.embed_dat` is called in other serializers inside `class_eval`, these two method calls get evaluated like other `attribute` calls in the serializers. I really don't understand this part of ruby well, but I eventually got it to work by using this code haha.

So in the end, for every key in the `embeddable` array, we're calling either attribute and has_many, or attribute twice (in the has_one case). You'll see in the options to the attribute call, I'm using the options passed by the `embeddable`. That means you can do `{ creator: { serializer: UserSerializer } }`, and it gets passed on to the attribute call here.

Ignoring pundit for a second, this is essentially like doing

```
class MappingSerializer < ...
  ...
  attribute :user_id, unless: -> { embeds.include?(:user) }
  attribute :user, if: -> { embeds.include?(:user) }
  attribute :map_id, unless: -> { embeds.include?(:map) }
  attribute :map, if: -> { embeds.include?(:map) }
end
```

This doesn't seem DRY, hence embed_dat.

Finally, I tried a few approached to this PR, but ultimately found out that attribute and has_many can take a block to change what's returned. This approach worked best, hence I added the block to has_many and introduced the policy_scope

PR #842, opened 2016-10-25
allow anonymous users to GET api routes

devvmh:
- /maps
- /maps/id
- /topics
- /topics/id
- /synapses
- /synapses/id


devvmh:
I think it did, but I didn't know that, so my tests weren't set up and there was at least one 500 error preventing it. Now the tests will help us understand that better and the docs will back it up

PR #844, opened 2016-10-26
fix topic card title word break css

### LINKED ISSUE #843: Really long words, or links, don't fit in the topic card
image

Should we hyphenate any words that we know will flow over the side? (i.e add a - and a space when displaying words over a certain limit)

The other problem would be, when the user goes to edit the actual title, we would have to show the true text... so the input box would still likely flow over the edge of the topic card.

devvmh:
fixes #843

### DIFF
--- a/app/assets/stylesheets/base.css.erb
+++ b/app/assets/stylesheets/base.css.erb
@@ -69,6 +69,7 @@
 }
 
 .CardOnGraph .title {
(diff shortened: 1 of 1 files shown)


devvmh:
screen shot 2016-10-26 at 8 41 38 am

here's a topic with proper word-break css

PR #854, opened 2016-10-26
Fix/pan.close.filterbox

Bortseb:


devvmh:
I'm still not sure I like this change. What if I want to pan around while filtering? It's kind of like pinning open the topic carousel - you leave it open and pan around while doing operations.

I think we should fix the click-pan issue instead

Bortseb:
Maybe this issue needs broader input then? To try and determine which option is more likely to be what the user expects... and to determine what is more likely, 1. that the user chooses some filters and then expects the box to go away once they resume mapping, or 2. That they more often like to keep the filter box open and pan around while they are adding filters. (And having to keep opening it is more annoying)

devvmh:
@Bortseb should we close this?

Bortseb:
sure, how things work now are good... i feel like we need a section for obscure options/tweaks for power users lol

PR #860, opened 2016-10-27
reorder API authentication precedence to be: token, oauth, cookie

devvmh:
raml is the best man

### DIFF
--- a/doc/api/apis/mappings.raml
+++ b/doc/api/apis/mappings.raml
@@ -1,7 +1,7 @@
 #type: collection
 get:
   is: [ embeddable: { embedFields: "user,map" }, orderable, pageable ]
-  securedBy: [ null, cookie, token, oauth_2_0 ]
+  securedBy: [ null, token, oauth_2_0, cookie ]
   responses:
     200:
       body:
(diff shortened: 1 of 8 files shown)


doc/api/apis/mappings.raml
```diff
@@ -1,6 +1,7 @@
 #type: collection
 get:
   is: [ embeddable: { embedFields: "user,map" }, orderable, pageable ]
+  securedBy: [ null, cookie, token, oauth_2_0 ]
```
Connoropolous: these ones say cookie, token, oauth2, and I think it should be the other way around?

devvmh:
I think I replied on Slack, but ultimately it's for developer happiness. If I go to the trouble of specifying an access_token in the url, I want that to be the user I'm logged in as. So if I'm developing with multiple users, but my browser happens to be opened as user1, I should be able to override that with user2's token

I'm actually still not 100% sure how doorkeeper works, does it use a query param too? Anyways, maybe doorkeeper/oauth2 should be first; we could change that if it should be. But I do think cookie should be last

PR #862, opened 2016-10-28
fix metacodes page scrolling

devvmh:
use scheme from
http://brandonhilkert.com/blog/page-specific-javascript-in-rails/


Connoropolous:
there's also metacode sets.

devvmh:
😂

PR #870, opened 2016-10-29
to_json should be as_json or there's bugs

devvmh:


Connoropolous:
surprised this wasn't causing more bugs. I am a bit worried about the confusion between the to_json and as_json, are we sure that changing it in the all the places won't break other tings?

devvmh:
I think the big reason it wasn't causing more bugs is that it only pops up if the json is within a larger structure, like

```
render json: @map.to_json
```

is fine, but

```
render json: { map: @map.to_json, topic: @topic.to_json }
```

is not

devvmh:
just deployed this to heroku for testing

PR #872, opened 2016-10-29
filter maps by user_id in api

devvmh:


doc/api/examples/map_starred.json
```diff
@@ -0,0 +1,28 @@
+{
+    "data": {
+        "id": 2,
+        "name": "Emergent Network Phenomena",
+        "desc": "Example map for the API",
+        "permission": "commons",
+        "screenshot": "https://s3.amazonaws.com/metamaps-assets/site/missing-map.png",
+        "starred": true,
+        "created_at": "2016-03-26T08:02:05.379Z",
+        "updated_at": "2016-03-27T07:20:18.047Z",
```
Connoropolous: this map doesn't have the user_id property on it, which is particularly confusing for the endpoint it's documenting I think :)

devvmh: dangit, rspec is supposed to catch this kind of problem, but i can't figure out how to make it enforce either one field or the other

PR #880, opened 2016-10-29
realtime topic and synapse delete were broken

### LINKED ISSUE #877: synapse or topic deleted by other during realtime not working
Is because of this:
https://github.com/metamaps/metamaps/blob/develop/frontend/src/Metamaps/Realtime/receivable.js#L45
and this:
https://github.com/metamaps/metamaps/blob/develop/frontend/src/Metamaps/Realtime/receivable.js#L108

error on heroku says: 
e.synapseRemoved is not a function

Connoropolous:
fixes #877


Connoropolous:
the reason for this is because the functions in Realtime/receivable get called directly by the socket events, and aren't added as functions to the global Realtime object.

Issue #833, opened 2016-10-24
Enable maps to be shared with the scaling and translation specified in the URL


Bortseb:
This would mean you can share your map to someone, and customize exactly what they would see on the map when they click the link.

Connoropolous:
the way to best accomplish this is likely to be the option of setting the pan and zoom values in the url, using `history.pushState` or `replaceState` , every time that the user finishes zooming or panning
https://css-tricks.com/using-the-html5-history-api/

then, to get the particular view they want, they just find that view themselves, and copy the url from the navbar

Connoropolous:
The second part of this issue involves grabbing those params off the queryString when the map loads, and zooming and panning accordingly. this would go into one of the `init` functions, we can figure out which module it makes sense to add it into

devvmh:
this is working, though a bit inaccurately, in #883

PR #883, opened 2016-10-30
enable specifying scale/translate in query params for a map

devvmh:
e.g. https://metamaps.cc/maps/12?scale=0.5&translate=-350,300

related to, but does not resolve, issue #833


frontend/test/Metamaps.Util.spec.js
```diff
@@ -125,4 +125,7 @@ describe('Metamaps.Util.js', function () {
   describe('resizeCanvas', function () {
     it.skip('TODO need a canvas')
   })
+  describe('queryParams', function () {
+    it.skiy('TODO need window')
+  })
```
Connoropolous: Needs to be `it.skip` (spelling)

devvmh:
I got it to update the query string when you zoom/pan the map in this PR!

Connoropolous:
this is exciting! would love to see this one out on heroku for testing soon as well. I've started reading through the code.

frontend/src/Metamaps/Visualize.js
```diff
@@ -152,8 +152,9 @@ const Visualize = {
 
     // monkey patch scale function
     const oldScale = self.mGraph.canvas.scale
+    const cachedPathname = window.location.pathname
     const updateScaleInUrl = debounce(() => {
-      Util.updateQueryParams({ scale: self.mGraph.canvas.scaleOffsetX.toFixed(2) })
+      Util.updateQueryParams({ scale: self.mGraph.canvas.scaleOffsetX.toFixed(2) }, pathname)
```
Connoropolous: isn't `pathname` undefined here?

frontend/src/Metamaps/Visualize.js
```diff
@@ -166,7 +167,7 @@ const Visualize = {
     const updateTranslateInUrl = debounce(() => {
       const newX = self.mGraph.canvas.translateOffsetX.toFixed(2)
       const newY = self.mGraph.canvas.translateOffsetY.toFixed(2)
-      Util.updateQueryParams({ translate: `${newX},${newY}` })
+      Util.updateQueryParams({ translate: `${newX},${newY}` }, pathname)
```
Connoropolous: and here?

frontend/src/Metamaps/Visualize.js
devvmh: thanks, fixed. it was working anyways since I had a default argument in Util, but fixing is good

Connoropolous:
I wonder about calling these `zoom` and `pan` in the url, just for easier readability / understanding by users, less engineer-y

Bortseb:
FWIW I like zoom more than scale (for similar reasons to connor)... But I prefer translate slightly more than pan because it is a math term, instead of a film (?) term... and we aren't really panning a camera :P

Not a big deal for me either way though

Connoropolous:
Isn't it called panning also when you mouse down, grab a surface and move
it? Like in Google maps?

Looks like Google has avoided the issue of naming these parameters. This is
kind of nice and is a behaviour that we could roughly model after
https://www.google.com/maps/@43.6711147,-79.3046103,17z

On Jan 15, 2017 11:59 AM, "Robert Best" <notifications@github.com> wrote:

FWIW I like zoom more than scale (for similar reasons to connor)... But I
prefer translate slightly more than pan because it is a math term, instead
of a film (?) term... and we aren't really panning a camera :P

Not a big deal for me either way though

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/metamaps/metamaps/pull/883#issuecomment-272708011>, or mute
the thread

Connoropolous:
@devvmh I'm going to work on this a little bit and come up with a final version of how to store it in the URL. This will be so useful to get in!

devvmh:
That's great! 

This is a big "public api" decision, so i think it's prudent to really think it through before pushing to metamaps.cc. once we enable it it'll be hard to turn off. So I'm okay if we merge,  but also ok if we wait

Connoropolous:
yeah, exactly. agreed. I want to design a good solution here

Bortseb:
Can't wait to have this one deployed! It has so many powerful uses and implications

PR #897, opened 2016-11-01
fix api docs

### LINKED ISSUE #887: api docs reference 'users/current' and 'user/current' and should both be the former

devvmh:
fixes #887

### DIFF
--- a/doc/production/pull-changes.md
+++ b/doc/production/pull-changes.md
@@ -24,9 +24,9 @@ Now that you have the code, run these commands:
 
     bundle install
     npm install
-    rake db:migrate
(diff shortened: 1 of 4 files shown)


doc/production/pull-changes.md
```diff
@@ -25,8 +25,7 @@ Now that you have the code, run these commands:
     bundle install
     npm install
     rake db:migrate
-    rake assets:precompile # includes `npm run build`
-    bin/build-apidocs.sh
+    rake assets:precompile # includes `npm run build` and `bin/build-apidocs.sh`
```
Connoropolous: Looks like its switching to `rails` so should it say 'rails assets:precompile' now?

PR #899, opened 2016-11-01
move webhook and event serializers back to top level

devvmh:


Connoropolous:
Why's this?

devvmh:
This is why the webhook service was failing

PR #902, opened 2016-11-01
fix db schema

devvmh:
I'm not sure what half of this means, but this will update the schema to match production

The next step, after merging this, is for me to run `rake db:schema:dump`, and then see how to update my local db to match the schema. That way when we merge db migrations, the same PR can include matching updates to db/schema.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -75,7 +75,6 @@ ActiveRecord::Schema.define(version: 20161105160340) do
     t.datetime "updated_at",    null: false
     t.integer  "mappable_id"
     t.string   "mappable_type"
-    t.boolean  "in_trash"
     t.index ["map_id", "synapse_id"], name: "index_mappings_on_map_id_and_synapse_id", using: :btree
     t.index ["map_id", "topic_id"], name: "index_mappings_on_map_id_and_topic_id", using: :btree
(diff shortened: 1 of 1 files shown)


devvmh:
Oh, look at that, it fails Travis. curioser and curioser. I"m going to investigate what's going on with this, I'd like our schema to be the same in all dev envs if possible.

Connoropolous:
Ohh, what happened here was experimental development going on in a branch on my C9 environment, and then the schema change must have got committed. Any recommendations for avoiding that in the future?

That's where the 'in_trash' column came from

devvmh:
I think this us just about code review.  if you see database changes in a
pp, the migrations should match the schema changes.

Honestly though,  the schema only causes problems for ci and  setting  up
new devs.

I'm more concerned  about  these weird additions to the table definitions.
I've never seen them before

devvmh:
UPDATE: ok, I renamed the postgresql "sequences" on linode, so now the only changes necessary are removing the in_trash columns. This should help us match our schemas to production.

If developers are on the develop branch with a clean DB and run `rake db:schema:dump`, it should make no changes to the schema file. If it does, the solution is to run `rake db:reset` (which will wipe all local data, but also makes the tables in line with the schema, which should now match production)

PR #915, opened 2016-11-05
Add Channel to Webhooks

Connoropolous:
so that the same webhook can be used multiple times to post to different channels


Connoropolous:
I tested this

PR #928, opened 2016-11-08
[develop] show private, shared maps in All Maps

devvmh:
map_scope already handles permissions, so we can remove the explicit filter on permission: 'private'

### DIFF
--- a/app/controllers/explore_controller.rb
+++ b/app/controllers/explore_controller.rb
@@ -9,7 +9,7 @@ class ExploreController < ApplicationController
 
   # GET /explore/active
   def active
-    @maps = map_scope(Map.where.not(name: 'Untitled Map').where.not(permission: 'private'))
+    @maps = map_scope(Map).where.not(name: 'Untitled Map')
 
     respond_to do |format|
       format.html do


Connoropolous:
My concern is that people see their own private maps in that list, and get worried that it shows up for everyone

Connoropolous:
That's why I set it that way

devvmh:
That's fair. I was actually going to comment that I found it strange that "my maps" private maps didn't show up in the list.

Maybe we can close this PR, and I'll think about implementing the configurable "start page", so I can load Metamaps and look at "my maps" instead of the "all maps" list.

Bortseb:
Maybe we show the permission of the map somewhere on the explore maps cards? (like we do on the map card on the map)

This would make it so that someone can quickly reassure themselves? But, they would likely only have to do this once or twice before they just trust the system (if they ever do it at all)

devvmh:
It's hard to reassure yourself without another account though (or a feature like Facebook has - "view as ____")

poietic:
yeah...i've found myself wanting permissions on the map card itself...but permissions is something we've got to sort out more now that we've got shared public and private maps, which also opens up shared public and private topics/synapses...i think its fine for now that its in a separate tab. but "All maps" is misleading then...

Bortseb:
for reassuance... I just mean, maybe show that it is indeed private as expected. And not something else...

knowing the precise/naunced permission (i.e. who exactly can edit this private map) etc.

devvmh:
I think I'm fine closing this. It could really go either way, but we can always change this later if we want.

PR #931, opened 2016-11-24
realtime shouldn't poll forever if the dev server isn't up

devvmh:


Connoropolous:
You tested this with the env set to both development and production?

devvmh:
Well, it behaves as I expected. I didn't deploy it or anything.

The real problem would come if we had long-running clients and we restarted the server. I think 5 (well, actually 6 because it does tries for 0, 1, ..., 5) is plenty though for this case

PR #932, opened 2016-11-25
mailboxer notification centre

devvmh:
This PR could be merged as is, except users might be confused about the difference between regular emails and internal notifications. The next step is to integrate all mailers to go via mailboxer.

Mailboxer is a great gem that's got lots of good database design behind it we can leverage. It's nice that it's modular and contained. In the future, we could integrate map and topic activity streams based on mailboxer. We could also set up user-to-user messaging, and potentially even store our text chat messages on each map in the mailboxer tables.

update 2 dec 11: feature is ready and live on heroku for testing. Merging soon.

update: new bugs and stuff

- [x] `devinhoward` requested access to `devin`'s map, and `devin` granted it. now when `devinhoward` visits the map, it's not marking the notification as read, perhaps because i did the notification wrong in the AccessRequest model file.
- [x] add pagination
- [x] unsubscribe link in email needs to be https, which doesn't seem to be true atm on heroku, might be fine on prod. 

~For now, I'm calling this WIP, until we do these things:~

- [x] integrate our current email notifications into mailboxer
- [x] text-based emails -> find a way to turn the html links into urls
- [x] get design feedback

I would also love it if we could do these things before merging, but don't consider them as blocking the merge:

- [x] set up our email links so when you take the actions (e.g. Grant/Deny for an access request), it marks the corresponding mailboxer notification as read
- [x] improve the text/html ability of the mailboxer/mailer integration (or go full-on mailboxer and lose the mailers)
- [x] add unsubscribe links to our emails
- [x] fix the bolding on the unread notifications
- [x] move the notification icon out of the account menu onto the main view
- [x] mark as read/mark as unread text
- [x] make the entire notification clickable
- [x] Back to Mapping -> remove !

Here is the current look and feel of the notification system as of noon on Nov 22:

index view
show view
desktop menu
account menu


Connoropolous:
Suuuuuper cool Devin. Some questions I have:

Where does 'back to mapping' take you?
Are you going to move the notifications icon out into plain view like Ishan suggested? (I'm for it)
Did you now get all the emails sending through mailboxer so they can be unsubscrived from? Is there a place within the platform where you can go directly to disable emails?

Comments:
Id suggest removing the exclamation point on 'back to mapping'. Reason being I think it just adds a level of cheesiness to what we're doing, when I'd like people to be able to have fun with it, but still understand that on some level it's a platform that means business.

I think it'd be preferable to make the entire notification, in list view, clickable. 

Instead of using brackets, using explicit phrasing for 'mark as read' and 'mark as unread' might be preferable. 
Let's unbold the main notification text?

Connoropolous:
Looking at GitHubs just reminded me of another thing which is a 'no unread notifications' message which can be kind of playful if we wqnt

devvmh:
github "deletes" notifications once you read them, but since ours are actionable it's probably better to leave them. So I'm not sure what it would look like. i guess we could have mark as read & delete. (i think a trash is even built into mailboxer).

I'll make those changes soon, they all seem good.

devvmh:
the emails are now all sending through mailboxer, but it's pulling extra header data from the existing mailers that is going to be tricky to remove. My last commit gets me a bit closer.

Back to mapping currently takes you to the root path `/`, which is just Explore Maps.

Yeah i'll move the notification icon out.

devvmh:
By "unbold the main notification text" you mean in the second screenshot above, right?

That's my bolding problem - that text is using the 'din-medium' font, and the font-weight is normal. I'm really not sure how to make our font not bold. Maybe I lower the font size???

Connoropolous:
Maybe the main notifs page just shows unread ones, and you can click 'archive' or 'see all' or something to view a list of all read notifs. 

I'll sort out how to unbold the text and commit it.

Connoropolous:
will you update screenshots? :) where does it position the 'notifs' icon when you're on a map page for example? what about on the 'explore' pages?

Connoropolous:
screen shot 2016-11-28 at 5 38 13 pm

screen shot 2016-11-28 at 5 38 38 pm

Connoropolous:
unbolding the text involves switching the font to 'din-regular' instead of 'din-medium' @devvmh

Connoropolous:
I did it in the commit that I made

Connoropolous:
I also made the icons, tooltip, and hover states work

devvmh:
😍  thanks

just gotta fix the mailer bodies so they don't include headers and we're good to go! I'll probably finish this tomorrow and we can do the final review

app/views/layouts/_upperelements.html.erb
```diff
@@ -72,7 +72,7 @@
   <% end %>
 
   <% if current_user.present? %>
-    <%= link_to notifications_path, target: '_blank', class: "notificationsIcon upperRightEl upperRightIcon #{user_has_unread_notifications? ? 'unread' : 'read'}" do %>
+    <%= link_to notifications_path, class: "notificationsIcon upperRightEl upperRightIcon #{user_has_unread_notifications? ? 'unread' : 'read'}" do %>
```
devvmh: no target _blank? Are you sure?

Connoropolous:
so is there anything that actually generates notifications right now?

devvmh:
yes check in access_controller, e.g.

https://github.com/metamaps/metamaps/pull/932/files#diff-1ba812c40e3c98ed9945f9c70d743e0cR25

I changed it so it uses the mailer to generate the HTML, and then does @map.user.notify (a mailboxer method added to the User model). Right now the integration is dumb, but it is actually sending stuff. I need to fix it (tomorrow)

Connoropolous:
so then, does the text of the notification just match the text of the email exactly? 
and does it even include the buttons?

devvmh:
That's my plan! That sounds desirable, right? I want the notification centre to encompass all mailings, so that unsubscribing from emails won't stop any messages from being sent.

Connoropolous:
so how would I generate a notification? I have the branch checked out

Connoropolous:
Is there paging for notifications? What would happen if a user accumulated a lot of them?

devvmh:
Good point. I'll add pagination when I'm hacking next

Connoropolous:
Overall, it is primarily styling and formatting that needs revision it would seem to me. 

A couple issues: 
screen shot 2016-12-11 at 9 56 56 pm
My current version of the tooltip can't handle text that changes length, in terms of its positioning, I'll need to do something to handle that. 

screen shot 2016-12-11 at 9 57 36 pm
Map links open in a new tab, when they should really open in the same tab instead. Perhaps even if we remove target=_blank it will still open in a blank target in email clients because it always does. That would solve this issue. 
Also it'd be best to ship it only if we get that bit of refactoring done that takes the normal 'email' formatting out and leaves only the stuff relevant to the notification in this context. 
The "back" div is a different width for some reason. 

screen shot 2016-12-11 at 9 57 03 pm
There's an empty div. Maybe its for pagination, but when there is nothing to paginate?

screen shot 2016-12-11 at 9 57 17 pm
There's a peculiar underlining that happens when you hover over the link for a notification. 
Styling also generally needs work here. 

Nevertheless, that is only speaking to 'what remains' and there's a tremendous amount of work here to celebrate, for what this opens up. You can bet that by having a system in place for generating notifications, we are going to start trying different ones out and opening up that dimension of the platform, and that's going to be sooo powerful. 🍑

Connoropolous:
When I just got an email from heroku the subject line started off "Mailboxer new notification - ..." so we'll want to change that as well

devvmh:
Hmmmmm I'll look into the subject problem

Connoropolous:
It would be good to give some more consideration to the "from" address of the emails that we send, and configuring that nicely. It doesn't have to happen right now, but I am just capturing it here cause I think it's important

devvmh:
I had set the "from" address to no-reply@metamaps.cc, but there was a verification error because someone (heroku? my gmail?) had determined that that mailbox didn't exist on the server. So potentially we're stuck with team@metamaps.cc unless we can figure out why that error happened. That's really hard to do because I don't have access to the email hosting setup

Connoropolous:
@devvmh I am thinking actually that the design work WITHIN the notifications themselves can wait. Because it will affect both how it looks in emails, and in the app, but for now the functionality is there, and it gets our legal "unsubscribe" requirement out sooner. So I am happy with this PR from a technical and design standpoint. =D you?

Connoropolous:
latest styles of the notifs list

screen shot 2016-12-15 at 11 52 12 pm
screen shot 2016-12-15 at 11 51 25 pm

PR #934, opened 2016-11-27
Active.Mapper was being initialized twice, causing errors

devvmh:


devvmh:
This is a bug because i added another line further  up in the file as part
of the  datamodel refactor. instead of being set directly in rails,  rails
sticks it  in serverData now

Connoropolous:
As long as at the end of the day Active.Mapper is still backbone initiated then it's all good :)

Issue #952, opened 2016-12-08
Various changes to how how topic/synapse creation process is cancelled or re-initiated


Bortseb:
1. Hitting ESC at any time should cancel whatever you are creating/changing (even if you are in the middle of Right-click drag and haven't even released your mouse button yet)
*Currently this only works when you are creating a synapse between two existing topics, and you hit ESC when prompted to "describe the connection"

2. Right-clicking & Right-dragging cancels topic creation and synapse creation. (If you are currently in the middle of creating a synapse, right-dragging on another topic on the map should re-initiate a whole new synapse creation originating from that new source)
*This doesn't happen in any current situation.

devvmh:
Done on develop

Issue #951, opened 2016-12-08
if i drag a synapse onto a topic and hold it there for a second or two, create the synapse (after showing a little affordance{


devvmh:

Connoropolous:
really loving this idea, especially the idea of a little progress bar that starts wrapping around the topic quickly till it completes and creates the synapse

Connoropolous:
related: https://github.com/metamaps/metamaps/issues/960

Bortseb:
I replied to this issue in our slack... but it seems possible that that might go unread... so I'll paraphrase much of it here

I prefer releasing the mouse button to confirm the creation of a synapse. There are many times when one might stop to think mid-way through synapse creation... and wouldn't want to create a bunch of junk synapses because they just happened to stop and think above another topic.

Also, for context... This idea came about because ben wants to make canceling the creation of a synapses be what happens when you "click away" instead of hitting enter during their creation.... and therefore that would totally eliminate the possibility of mouse-only interaction. (this issue came about as a solution to that problem from what I recall)

I prefer things how they are now, because I like being able to quickly create synapses without having to name them just yet.

Also, it is hard to evaluate this branch, because I can't currently create topics on heroku... also hovering over a topic during synapse creation just keep creating more synapses for me (like, it stacks them until I move my mouse again) also, it makes it so that it creates a synapse to other topics simply by hovering over another metacode, regardless of the time period, so I need to dodge other topics if I want to avoid creating a synapse to it. (on my way to creating the synapse I do want to create)

devvmh:
^^ to the above, I think I need to cancel the timeoutId in more branches of the click-drag function; that way it won't create the synapse if you roll over the topic for a second

PR #956, opened 2016-12-08
Feature/overhaul synapse creation

### LINKED ISSUE #949: fix synapse creation bugs
1. Clicking away from incomplete synapse label should cancel it not confirm it
2. Backspace to empty field while typing label should maintain creation dialog rather than canceling

devvmh:
fixes #949, #951, #952. 

We could pick and choose from the 4 commits included.

bugs:

- [ ] create a new topic, hit enter, and it says please enter topic title (robert found this one)


devvmh:
@Bortseb I can't replicate your error where it says "Please enter topic title" when you hit enter. can you give more details?

Bortseb:
I double click to create a topic, input my title.... then hit enter.. However, the topic doesn't get created:

image

Connoropolous:
I think the delay before synapse creation happens automatically needs to be a bit longer, and also that we do need the visual indicator of some kind before it happens. 
Small bug with it at the moment is that the topic the synapse was created to remains enlarged after the synapse is created. 
screen shot 2016-12-08 at 5 12 50 pm

devvmh:
ok i believe all bugs are fixed in the latest (deployed to heroku) except for the problem where you roll over and then move away

devvmh:
I've reworked this to 6 commits. I'm going to post the first 4 as a standalone PR https://github.com/metamaps/metamaps/pull/970 (esc or right click cancel, tab confirms). the other two will be what we're debating in this PR (left click cancels synapse creation, and we mitigate the fallout from that by allowing drag and hover to create synapses)

Issue #969, opened 2016-12-10
(130) Make DEL key delete selected topics and/or synapses


Bortseb:
keeping our existing confirmation prompt of course.

This is so that we can eliminate our current CTRL+D, which conflicts with many browser's bookmark shortcut.

Do all ecosystems have a DEL key on their keyboard?

============
100BD/C = 100*1*0.65/0.5= 130

devvmh:
Mac keyboards dont have del. Their backspace is labelled del, but it's the same as backspace.  I can hit Fn+Backspace to get the same effect as del, but this is too obscure to rely on. 

I would be a big fan of changing Ctrl+D to Alt+D, and allowing the use of del as well.

Bortseb:
lol, there are so many shortcuts!...
ALT+D is one I am more used to using because in both chrome and windows explorer it puts your cursor to the address bar

Bortseb:
We could make Backspace delete too... 

but some people might actually use that for going back to the previous browser page... testing it now though... we have already eliminated this on Metamaps.

Huh, so it seems Chrome must have changed this... when I hit backspace a bunch of times, below is what it shows me... so I think both DEL and BACKSPACE would be good options for us. Lets do them both for deleting stuff.

![image](https://cloud.githubusercontent.com/assets/1617130/21075906/dd9a8bc8-beeb-11e6-9862-22610a298d29.png)

devvmh:
Ugh you're right, Alt+D is used.

I think Backspace is pretty well established as a paradigm for going back. This Chrome thing must be quite new, since it's not on my browser yet. It's as least as strong as Ctrl+D for bookmarks. I'd want to see a stronger precedent from the browsers before changing the keyboard shortcut; it's never good to surprise our users if we don't need to IMO

Bortseb:
ok, so we can make DEL delete... this takes care of windows users... what do MAC/Linux normally use for delete? (Since you said their DEL/backspace is already used in browser for going back... what do other apps use for Delete?)

Bortseb:
I'm going to just make DEL work as described without any other changes... I try and use this every time I map. :P

Bortseb:
This issue has been resolved... along with resulting bugs from here:
#1083

PR #971, opened 2016-12-12
Made DEL key an alternative for deleting selected entities

Bortseb:


Bortseb:
see #969 for context

Connoropolous:
hm, I didn't get notified of this request for review. interesting. 

I will check it out locally on my computer, because the code looks fine, but I want to make sure I/we aren't missing anything. Nice Rob!

Bortseb:
Maybe you weren't notified @Connoropolous because I didn't add you at the time of creation... but instead added you after once a few days went by and I wasn't sure if anyone saw this little addition.... So maybe pings like this one are better?

Connoropolous:
@Bortseb in the cases of this proposal going through, lets change this to perform the `remove` action instead of `delete`
https://metamaps.slack.com/archives/metamaps-cc/p1481736452000162

So let's wait a little longer before merging

PR #984, opened 2016-12-14
Track everything we need to reconstruct maps

Connoropolous:
An important thing to note here is how it is captured with every event who the 'actor' was. 

Assuming that these changes make it in, the most glaring set of metadata we would not be logging are the edit history of topics and synapses. As it stands, every time someone makes a change to those values, we lose the old values. If we started to also capture those values, possibly also as `events`, then we would truly be able to reconstruct the lifecycle of a map, by merging the event streams of the map, with the event streams of the topics and synapses that were present on it at the time

### DIFF
--- a/app/controllers/api/v2/mappings_controller.rb
+++ b/app/controllers/api/v2/mappings_controller.rb
@@ -5,6 +5,21 @@ module Api
       def searchable_columns
         []
       end
+
+      def update
+        # mappings are the only things where the user is set to the latest updater
+        # done this way so that the model hooks can use the mapping user to determine who took this action
+        resource.user = current_user if current_user.present? # current_user should always be present
+        update_action
+        respond_with_resource
+      end
+
+      def destroy
+        # this is done so that the model hooks can use the mapping user to determine who took this action
+        resource.user = current_user if current_user.present? # current_user should always be present
+        destroy_action
+        head :no_content
+      end
     end
   end
 end
--- a/app/controllers/api/v2/mappings_controller.rb
+++ b/app/controllers/api/v2/mappings_controller.rb
@@ -6,19 +6,23 @@ module Api
         []
       end
 
+      def create
+        instantiate_resource
+        resource.user = current_user if current_user.present?
+        resource.updated_by = current_user if current_user.present?
+        authorize resource
+        create_action
+        respond_with_resource
+      end
+
       def update
-        # hack: set the user temporarily so the model hook can reference it, then set it back
-        temp = resource.user
-        resource.user = current_user
+        resource.updated_by = current_user if current_user.present?
         update_action
         respond_with_resource
-        resourse.user = temp
-        update_action
       end
 
       def destroy
-        # this is done so that the model hooks can use the mapping user to determine who took this action
-        resource.user = current_user if current_user.present? # current_user should always be present
+        resource.updated_by = current_user if current_user.present?
         destroy_action
         head :no_content
       end
--- a/app/controllers/api/v2/mappings_controller.rb
+++ b/app/controllers/api/v2/mappings_controller.rb
@@ -7,11 +7,13 @@ module Api
       end
 
       def update
-        # mappings are the only things where the user is set to the latest updater
-        # done this way so that the model hooks can use the mapping user to determine who took this action
-        resource.user = current_user if current_user.present? # current_user should always be present
+        # hack: set the user temporarily so the model hook can reference it, then set it back
+        temp = resource.user
+        resource.user = current_user
         update_action
         respond_with_resource
+        resourse.user = temp
+        update_action
       end
 
       def destroy
--- a/app/controllers/mappings_controller.rb
+++ b/app/controllers/mappings_controller.rb
@@ -31,6 +31,8 @@ class MappingsController < ApplicationController
   def update
     @mapping = Mapping.find(params[:id])
     authorize @mapping
+    # hack: set the user temporarily so that the model hook can reference it, and then set it back
+    temp = @mapping.user 
     @mapping.user = current_user
     @mapping.assign_attributes(mapping_params)
 
@@ -39,12 +41,16 @@ class MappingsController < ApplicationController
     else
       render json: @mapping.errors, status: :unprocessable_entity
     end
+    # restore the original mapping creator
+    @mapping.user = temp
+    @mapping.save
   end
 
   # DELETE /mappings/1.json
   def destroy
     @mapping = Mapping.find(params[:id])
     authorize @mapping
+    # hack: set the user temporarily so that the model hook can reference this user who is taking the action
     @mapping.user = current_user
     @mapping.destroy
 
--- a/app/controllers/mappings_controller.rb
+++ b/app/controllers/mappings_controller.rb
@@ -22,7 +22,6 @@ class MappingsController < ApplicationController
 
     if @mapping.save
       render json: @mapping, status: :created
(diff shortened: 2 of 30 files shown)


app/models/event.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 class Event < ApplicationRecord
-  KINDS = %w(user_present_on_map conversation_started_on_map topic_added_to_map synapse_added_to_map).freeze
+  KINDS = %w(user_present_on_map conversation_started_on_map 
+               topic_added_to_map topic_moved_on_map topic_removed_from_map
+               synapse_added_to_map synapse_removed_from_map).freeze
```
Connoropolous: By capturing this particular set of events:
```
topic_added_to_map
topic_moved_on_map
topic_removed_from_map
synapse_added_to_map
synapse_removed_from_map
```
and storing the x and y values of the topic at the time of creation, and then every time it moves, we have everything we need to reconstruct the current state of the map, and watch its evolution over time.

app/models/mapping.rb
```diff
@@ -16,11 +16,45 @@ class Mapping < ApplicationRecord
 
   delegate :name, to: :user, prefix: true
 
+  after_create :after_created
+  after_update :after_updated
+  before_destroy :before_destroyed
+
```
Connoropolous: By moving the event publishing to these ActiveRecord hooks in the model file, we ensure that they will always get called, regardless of which controller or other code takes the `create` `update` or `destroy` actions

db/migrate/20161214140124_add_meta_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddMetaToEvents < ActiveRecord::Migration[5.0]
+  def change
+    add_column :events, :meta, :json
```
Connoropolous: This new column can be used in diverse ways, depending on the event type. The only thing to be VERY careful of is consistency.

Connoropolous:
@devvmh all this needs is a technical review, and we need to remember to run the related migration on deployment. also, I didn't check in the schema.rb file cause I still had a bunch of weird stuff that messed around with it. maybe I need to drop my db and do a clean schema import. but either way that schema.rb update will need to get in here somehow

Connoropolous:
I am thinking of following this up extremely soon with the capturing of events related to topic and synapse updates

Connoropolous:
@devvmh I am not totally sold on the approach I'm taking in this commit... 
https://github.com/metamaps/metamaps/pull/984/commits/a6bb688d0cd1a1b71987db82907fc25dd55733a4

It doesn't keep a very granular record of what changed, the work for a developer to figure out what changed in the code, by comparing two of these records, would be reasonably high compared to if I made the event type more granular. Like: 'topic_name_updated' or 'synapse_desc_updated'
but that also generates a lot of code and I'm not sure if we need all that.

Connoropolous:
If you're not sure either, but you like the rest of it, that one could come off and I could take more time with it

devvmh:
Hmm. That commit seems pretty smart. The simple approach would be to store the whole topic/synapse on every commit, but storing what changed seems pretty smart too. It does make it harder to reverse if you only store what changed. But other than that it seems fine.

Really we need to know what our storage limits are. I guess this change will have us running into our heroku storage limits a lot more often, right?

This approach but storing the new & old values seems like the best approach.

Anyways, I'm fine with all of the approaches. They all seem good.

devvmh:
Can you run rubocop on the changed files? =D

app/controllers/api/v2/mappings_controller.rb
```diff
@@ -5,6 +5,23 @@ class MappingsController < RestfulController
...
       end
+
+      def update
+        # hack: set the user temporarily so the model hook can reference it, then set it back
+        temp = resource.user
+        resource.user = current_user
+        update_action
+        respond_with_resource
+        resourse.user = temp
+        update_action
```
devvmh: I'd really like to find a better way to do this

app/controllers/mappings_controller.rb
```diff
@@ -32,26 +31,27 @@ def create
   def update
     @mapping = Mapping.find(params[:id])
     authorize @mapping
+    # hack: set the user temporarily so that the model hook can reference it, and then set it back
+    temp = @mapping.user 
+    @mapping.user = current_user
+    @mapping.assign_attributes(mapping_params)
```
devvmh: what's the difference between update_attributes and assign_attributes?

app/controllers/mappings_controller.rb
```diff
@@ -32,26 +31,27 @@ def create
...
-
-    mappable = @mapping.mappable
-    if mappable.defer_to_map
-      mappable.permission = mappable.defer_to_map.permission
-      mappable.defer_to_map_id = nil
-      mappable.save
-    end
-
+    # hack: set the user temporarily so that the model hook can reference this user who is taking the action
+    @mapping.user = current_user
```
devvmh: I'm confused about why the following lines were removed, but also about why they were there in the first place. Any insights?

app/models/webhooks/slack/synapse_removed_from_map.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+class Webhooks::Slack::SynapseRemovedFromMap < Webhooks::Slack::Base
+  def text
+    connector = eventable.desc.empty? ? '->' : eventable.desc
+    # todo express correct directionality of arrows when desc is empty
+    "\"*#{eventable.topic1.name}* #{connector} *#{eventable.topic2.name}*\" was removed by *#{event.user.name}* as a connection from the map *#{view_map_on_metamaps}*"
+  end
+end
```
devvmh: are all these new webhooks configurable? or will they all start showing up on slack?

app/controllers/api/v2/mappings_controller.rb
Connoropolous: We could add an updated_by column to mappings, and then reference it in the model hook

app/controllers/mappings_controller.rb
Connoropolous: They got moved into the before_destroyed model hook. They make it so that if the topic or synapse being removed from a map was deferring to it for its permissions, set the permission one last time but then disassociate them

app/models/webhooks/slack/synapse_removed_from_map.rb
Connoropolous: Every webhook we set up in the db subscribes to whichever events out of all of these that it wants. So it's configurable. You could listen for all events, or just one

Connoropolous:
what are some tips for working with rubocop? I can see it output all the errors from the whole app, but I don't know how to work with this subset?

what do I do with a string line that is too long?

Connoropolous:
@devvmh I found a solution I'm happy with: just store alongside the attribute names that changed at this time
https://github.com/metamaps/metamaps/pull/984/commits/223d31c4f4cafb3dd12cbc7c1dc3b58ef89b61bf

app/controllers/api/v2/mappings_controller.rb
Connoropolous: alternatively, would it be ok to set a value temporarily on the model, because we don't actually need to persist it to the database, on the mapping model itself?

Connoropolous:
leaving another note to self/ @devvmh that when we deploy, that would be a good time to check in the schema.rb changes after we run the migration

Connoropolous:
this is ready to merge

Issue #193, opened 2013-04-02
Random floating topics (and other garbage data)


Bortseb:
Let’s say I create a topic (with a synapse) on a map, and later come back and decide I want to delete it. However I don't choose to permanently delete it because I have no idea if others have added it to one of their maps, so I just choose to remove it from the map I am working on. The issue I have found is that even though I have removed this topic from this map, and it is actually not part of any map any more, the removed topic is still there when I go to topic view. This make me wonder if topics created without a synapse, which are then later removed from their map, are also floating around in the database not belonging to any map. If you remove a topic from a map, should the system not determine whether that topic still exists on another map, and if it doesn't the propose to you that it be permanently deleted? This issue is an extension of #192 , and could be aided by the implementation of #188

Issue #192, opened 2013-04-02
Deleted synapses still show up in topic view


Bortseb:
Synapses connecting topics show up in the topic views even when they are no longer part of any maps (including the one the synapse was created on)

To see what I mean:
Try creating a new topic with a synapse. Then delete just the synapse. Then go to topic view for either of the previously connected topic.

Issue #163, opened 2013-03-15
Allowing users to permanenetly delete topics is dangerous


Bortseb:
This is especially true when topics were selected from the auto-complete drop-downs and are inside other maps. When is the use-case of needing to be able to delete all instances of a topic in metamaps?

The warning pop-up is nice, but maybe include in it all the other maps that the topic being deleted is part of?

Issue #152, opened 2013-03-01
Replace "Permanently delete selected" with "Trash selected".


poietik:
we shouldn't delete from the system - it should go into a trash that we can restore from.

PR #985, opened 2016-12-14
disable permanent deletion of topics and synapses, except...

Connoropolous:
when the topic or synapse no longer appears on any maps, and this can currently only be performed over the api

proposal was:
*We do a dead stop, for now, on full topic and synapse deletion from the database (unless they appear on no other maps, but we only leave that option available on the api, not the GUI). We remove it as an option.* 
Elaboration: our current implementation poses real risks to the end user. While they CAN only delete topics that they created, they CAN delete topics that appear on others maps, or other maps of their own, accidentally, easily. My plan would be two-fold: remove 'delete' as an option in the right click, and we could change the text of the other to just 'remove'. Then, we lock the server endpoints so that topics and synapses only successfully delete if they DONT appear on any maps. That is for security purposes. Third step is: design and develop a page for 'orphaned' topics and synapses, where they can be reviewed and either sent/added to maps, or permanently deleted. We can build this feature at our leisure because there is no detriment to the end user in the meantime. Please use the thumbs up or thumbs down to vote on this change. I will be asking for thumbs down votes to please clarify their position.

this was agreed upon here: 
https://metamaps.slack.com/archives/metamaps-cc/p1481736452000162

Looks like: 
screen shot 2016-12-14 at 2 55 37 pm

Pitfalls:
topics will still show up in autocomplete and search results, even when the end user might not expect them to. we have an option to hide results for the time being, if the topic or synapse in question doesn't appear in any maps, but that actually seems like a bad idea. because one could intentionally remove it from one map, and then want to use autocomplete to pull it into a different map. I think this issue will be cleared up when there's a "manage orphaned info" dashboard

### DIFF
--- a/app/policies/synapse_policy.rb
+++ b/app/policies/synapse_policy.rb
@@ -38,7 +38,7 @@ class SynapsePolicy < ApplicationPolicy
   end
 
   def destroy?
-    record.user == user || admin_override
+    record.mappings.empty? && (record.user == user || admin_override)
   end
 
   # Helpers
--- a/app/policies/topic_policy.rb
+++ b/app/policies/topic_policy.rb
@@ -36,7 +36,7 @@ class TopicPolicy < ApplicationPolicy
   end
 
   def destroy?
-    record.user == user || admin_override
+    record.mappings.empty? && (record.user == user || admin_override)
   end
 
   def autocomplete_topic?
--- a/frontend/src/Metamaps/Control.js
+++ b/frontend/src/Metamaps/Control.js
@@ -39,76 +39,13 @@ const Control = {
     Selected.Nodes.splice(
       Selected.Nodes.indexOf(node), 1)
   },
-  deleteSelected: function() {
-    if (!Active.Map) return
-
-    var n = Selected.Nodes.length
-    var e = Selected.Edges.length
-    var ntext = n === 1 ? '1 topic' : n + ' topics'
-    var etext = e === 1 ? '1 synapse' : e + ' synapses'
-
-    var authorized = Active.Map.authorizeToEdit(Active.Mapper)
-
-    if (!authorized) {
-      GlobalUI.notifyUser('Cannot edit Public map.')
-      return
-    }
-
-    var r = window.confirm(outdent`
-      You have ${ntext} and ${etext} selected. Are you sure you want
-      to permanently delete them all? This will remove them from all
-      maps they appear on.`)
-    if (r) {
-      Control.deleteSelectedEdges()
-      Control.deleteSelectedNodes()
-    }
-
-    if (DataModel.Topics.length === 0) {
-      GlobalUI.showDiv('#instructions')
-    }
+  removeSelected: function() {
+    Control.removeSelectedEdges()
+    Control.removeSelectedNodes()
   },
-  deleteSelectedNodes: function() { // refers to deleting topics permanently
-    if (!Active.Map) return
-
-    var authorized = Active.Map.authorizeToEdit(Active.Mapper)
-
-    if (!authorized) {
-      GlobalUI.notifyUser('Cannot edit Public map.')
-      return
-    }
-
-    var l = Selected.Nodes.length
-    for (var i = l - 1; i >= 0; i -= 1) {
-      var node = Selected.Nodes[i]
-      Control.deleteNode(node.id)
-    }
-  },
-  deleteNode: function(nodeid) { // refers to deleting topics permanently
-    if (!Active.Map) return
-
-    var authorized = Active.Map.authorizeToEdit(Active.Mapper)
-
-    if (!authorized) {
-      GlobalUI.notifyUser('Cannot edit Public map.')
-      return
-    }
-
-    var node = Visualize.mGraph.graph.getNode(nodeid)
-    var topic = node.getData('topic')
-
-    var permToDelete = Active.Mapper.id === topic.get('user_id') || Active.Mapper.get('admin')
-    if (permToDelete) {
-      var mappableid = topic.id
-      var mapping = node.getData('mapping')
-      topic.destroy()
-      DataModel.Mappings.remove(mapping)
-      $(document).trigger(JIT.events.deleteTopic, [{
-        mappableid: mappableid
-      }])
-      Control.hideNode(nodeid)
-    } else {
-      GlobalUI.notifyUser('Only topics you created can be deleted')
-    }
+  hideSelected: function() {
+    Control.hideSelectedEdges()
+    Control.hideSelectedNodes()
   },
   removeSelectedNodes: function() { // refers to removing topics permanently from a map
     if (Active.Topic) {
@@ -237,59 +174,6 @@ const Control = {
     Selected.Edges.splice(
       Selected.Edges.indexOf(edge), 1)
   },
-  deleteSelectedEdges: function() { // refers to deleting topics permanently
-    if (!Active.Map) return
-
-    var authorized = Active.Map.authorizeToEdit(Active.Mapper)
-
-    if (!authorized) {
-      GlobalUI.notifyUser('Cannot edit Public map.')
-      return
-    }
-
-    const l = Selected.Edges.length
-    for (let i = l - 1; i >= 0; i -= 1) {
-      const edge = Selected.Edges[i]
-      Control.deleteEdge(edge)
-    }
-  },
-  deleteEdge: function(edge) {
-    if (!Active.Map) return
-
-    var authorized = Active.Map.authorizeToEdit(Active.Mapper)
-
-    if (!authorized) {
-      GlobalUI.notifyUser('Cannot edit Public map.')
-      return
-    }
-
-    var index = edge.getData('displayIndex') ? edge.getData('displayIndex') : 0
-
-    var synapse = edge.getData('synapses')[index]
-    var mapping = edge.getData('mappings')[index]
-
-    var permToDelete = Active.Mapper.id === synapse.get('user_id') || Active.Mapper.get('admin')
-    if (permToDelete) {
-      if (edge.getData('synapses').length - 1 === 0) {
-        Control.hideEdge(edge)
-      }
-      var mappableid = synapse.id
-      synapse.destroy()
-
-      // the server will destroy the mapping, we just need to remove it here
-      DataModel.Mappings.remove(mapping)
-      edge.getData('mappings').splice(index, 1)
-      edge.getData('synapses').splice(index, 1)
-      if (edge.getData('displayIndex')) {
-        delete edge.data.$displayIndex
-      }
-      $(document).trigger(JIT.events.deleteSynapse, [{
-        mappableid: mappableid
-      }])
-    } else {
-      GlobalUI.notifyUser('Only synapses you created can be deleted')
-    }
-  },
   removeSelectedEdges: function() {
     // Topic view is handled by removeSelectedNodes
     if (!Active.Map) return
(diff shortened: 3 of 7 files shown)


Connoropolous:
I left Ctrl-M also working, for the removal of topics and synapses from a map, because that's what it was before.

app/policies/synapse_policy.rb
```diff
@@ -38,7 +38,7 @@ def update?
   end
 
   def destroy?
-    record.user == user || admin_override
+    record.mappings.empty? && (record.user == user || admin_override)
```
Connoropolous: maybe we should add some custom text as a disclaimer about this in the API documentation?

Connoropolous:
Update: on hold while we await Ben's comments.

Connoropolous:
Pulled from the slack chat and edited for relevance::

poietic [12:42 PM] 
okay. with that successfully delete if they don’t appear on any other maps makes sense to me. I would still like to have admin privileges to delete other people’s topics and synapses - I never use this except when i’ve been coordinating creating maps across accounts for specific purposes, and they’re on accounts that I’ve set up for others.

connor [12:43 PM]  
you would just be able to use ‘remove from map’ and wait until the ‘topic and synapse cleanup’ page got released to quickly scan through it and tidy up the database

[12:44]  
although that doesn’t solve the ‘admin can delete topics and synapses created by someone else’ problem

robert.best [12:47 PM]  
I like the idea... where my mind took this is that, if we are tracking the evolutionary progression of each map (so that we have everything we need to reconstruct it), then we could also do the same with a tree structure for topics too... So the point on the "map-tree" when a specific topic got added intersects with a time-stamp on the "topic-tree" when that topic got added...

This way if someone decides down the road that they want to permanently delete a topic, and remove it from all maps... i.e. make it an orphan... then the "owners" of the maps that this topic was on can decide to "adopt" the ophan, and just make a duplicate copy of it of their own using the state of the topic at the point that the topic-tree intersected with the map-tree...

Yep! steered right clear of that overthinking bit :stuck_out_tongue:

connor [1:02 PM]  
Heads up that I just enabled the wiki on the Metamaps repo, so that we could keep revision history of our ‘Primer on Metamaps Design’ along with other stuff like that
https://github.com/metamaps/metamaps/wiki/Primer-on-Metamaps-Design (edited)

[1:03]  
to your points Rob, there is a lot of relevance to that kind of thinking. I think you are really hitting on a good direction

[2:51]  
Stopping deletion sounds good to me! Including the events pr, this will be a big moment!

robert.best [2:56 PM]  
I find it interesting that a graph structure, specifically a https://en.wikipedia.org/wiki/Directed_acyclic_graph, would be great for representing and containing the entire evolutionary history of a map and its topics... so basically a metamap would eat another metamap in order to produce the final state of a "current" map...

This might not be all that useful to anyone right now... more just interesting. :stuck_out_tongue: (edited)

----- Yesterday December 15th, 2016 -----
benb [8:42 AM]  
Okay I think ees workinger now - my experience is there's three ways to deal with unwanted, obsolete, or error topics. Edit, delete, or orphan. I've used all three, with conscious intent

[8:45]  
And I like the ability to choose by case. Deleting is the fastest, cleanest solution in many cases, the only liability being inadvertent removal from other maps. Hence, the warning dialogue proposal for that specific situation

[8:47]  
The idea of adding on an extra workload of cross checking orphaned topics, and forcing people to do something slightly counterintuitive when they want to drop something from system, is poor design methinks

[8:50]  
It's something to revisit when we get to point of higher power background processing that can parse topics and intelligently amend or assign via issues/prompts maybe

connor [8:53 AM]  
I agree this would be great: "Hence, the warning dialogue proposal for that specific situation”
_and_ its really hard to implement. it could be something we come back to, in the meantime, it seems like there’s more harm than good may come of having permanent topic and synapse deletion.

benb [8:53 AM]  
Some big overall design Qs to revisit about how we want info to live in the graph, how to improve s/n, personalization, cognitive assist. Overall I'm in favor of minimizing redundancy of topic creation and senseless proliferation, but right now users seem to be doing that by nature or for efficiency and it's working, for them, for now

[8:54]  
Can you say what in particular is the obstacle to implement the warning?

[8:56]  
Hard to query that info, or figure out compound permissions, or?

connor [8:56 AM]  
well, it’s a complex user experience for one… lets say there were 10 topics and 5 synapses you wanted to delete, and turns out 5 of the topics and 2 of the synapses were on other maps. Do you have to respond to 7 separate prompts and warnings?

benb [8:58 AM]  
Yeah, I don't know that it would have to be so comprehensive. Just the numbers, i.e. 5 topics and 2 synapses are used on other maps: delete the others, delete all, delete none? (edited)

[8:59]  
I think just that much would give people the jheadsup they need.

connor [9:00 AM]  
what about the case where a topic doesn't appear on any other maps, but has a multitude of synapses connected to it

[9:01]  
"5 topics and 2 synapses are used on other maps: delete the others, delete all, delete none?” something like this could work, the other layer that you named which is true, is that the querying is rather hard

benb [9:01 AM]  
That's a bit stickier, but could tweak the warning wording to include those as well, like 6 topics are in use elsewhere in the system

[9:02]  
Can nudge people to open topic cards and check usage info as a reminder too, on warning blurb

connor [9:02 AM]  
we could write a custom endpoint to return the basic info needed to generate the warning. but the user will have to wait for a server response before being able to complete their action

[9:03]  
and even still, my intention would be to keep the actual record in the database, but add some kind of marker to it that indicates it as having been ‘trashed’ and thus not showing up in search results or the autocomplete

[9:04]  
how would it be if we did those things Ben?

benb [9:05 AM]  
Ahah, trash sounds slightly different, but still ends up dumping it on someone else to process and sort/clear if they're no longer available to users

[9:07]  
I'll think it over some more, but leaning towards simple incremental change there until we're  ready to update the paradigm more holistically

connor [9:07 AM]  
the end user sees the same thing, it’s just what happens in the database that’s different.

benb [9:08 AM]  
I'll take a look at the roadmap as well. Haven't been at a good enoug isp to get on maps for a few days

connor [9:08 AM]  
and there’s no harm in not permanently data that’s been generated at this point, because its irretrievable by anyone, for any purpose

[9:09]  
there’s no reversing it, and that’s the problem

benb [9:09 AM]  
No, end user decides about keeping or dropping topics from the graph - is more of an agent of the system and collective Intel that way I think. Want to reinforce that sense

connor [9:09 AM]  
we could run a script if we want, once a day, that cleans up stuff that’s been ‘deleted’ from the database that’s older than say 30 days old, for example

[9:10]  
not saying we would, but as an example

benb [9:12 AM]  
I'm thinking it's important to get that through to people, repeatedly, that this is a knowl commons of which all are stewards. It's a learning process, and retraining of old habit. Good, but slow and a bit awkward or frustrating

connor [9:13 AM]  
ok, well I will form an updated proposal and put it out

benb [9:13 AM]  
There's only so much offloading of those kinds of processes and discretion we'll be able or wanting to do eventually, in my mind. Crafting and curation has got to be somewhat hands on, want to make that participatory and equalized best we can

[9:14]  
But I get that there could be certain roles and privs for those specific tasks eventually, too

[9:17]  
I want all users to be conscientious of it anyhow, taking into consideration, being reminded

robert.best [9:19 AM]  
Given that something public/commons can be copied into anyone else's map... It doesn't make sense to me that you should be able to delete it from other's maps... it is effectively no longer just your own. And yea, we also don't give the user a great way to make the decision of wether or not its a good idea to go ahead and delete a topic. Seeing a # of maps that something is on doesn't really give much insight on the impact of the delete.

[9:20]  
actually, saying that... it makes me wonder should there be a difference in my ability to add public topics to my maps vs commons ones? Should another's public topic, which is meant to be "view only" in some sense, be able to be added to my map? (edited)

devvmh [1:00 PM]  
I think i like what Connor proposed better. Even if I'm totally informed and enthusiastic about differentiating deletion and removal from a map,  I'll always have this nagging suspicion that a topic might be on another map (or accidentally delete a topic from someone else's map)  And if im uninformed or lazy or malicious,  I'll end up deleting topics from others' maps.

I think there are better ways to educate the user about the knowledge graph than allowing lots of deletion

[1:01]  
Maybe there could be an option in the topic card that would do a forced deletion from all maps.  It would feel like a bigger decision then,  whereas a keyboard shortcut is easy to use and abuse

benb [3:44 PM]  
How much deletion is actually happening thus far? And how much that's been problematic? I can think of two cases where stuff was wiped off maps and it was a drag. Other instances where commons topics have been edited beyond recognition of original intent. I bet all of the above would have been averted with appropriate cautions

[3:46]  
On the other hand I definitely delete topics and synapses regularly, after checking whether they're in use elsewhere. Often it's stuff I,ev recently created that I realize I don't want, and actually don't want in the system at all. Especially synapses

connor [3:48 PM]  
"How much deletion is actually happening thus far?” we don’t know :stuck_out_tongue: there’s no trace of it

benb [3:48 PM]  
Footnote about starting to play with a time slider kind of feature - it occurred to me recently to introduce that into the filter menu, to filter by time of creation on map. Perhaps an auto save increment at first

connor [3:50 PM]  
I was playing with some code this morning, alongside the backend changes that I’ve been working on, that is around this idea of playing back the development of the map.

benb [3:50 PM]  
Ahkay. Perhaps we should upgrade from 'obliterat' to just regular old delete then, in prep for allowing undo etc

connor [3:51 PM]  
that’s the idea ^ :slightly_smiling_face:

robert.best [3:51 PM]  
I A special case I would be ok with is deleting anything that you created, that is only on your own maps

[3:55]  
but if we are keeping track of everything, so as to be able to undo it... like as just discussed... then there may be no need...

Even if we don't call it this... once something is removed everywhere, such that it only resides as a potential to be "undone" then it seems to be "archived"... But maybe  there will still be times when we want to take this one step further and remove all traces... (though, we aren't enabling that in the near future it doesn't seem)

devvmh [6:13 PM]  
Whatever we change,  I'm hoping it simplifies the end user ux and makes deletion vs removal intuitive,  instead of making it more confusing

[6:14]  
That is, i hope we end up having one prominent option (remove), and either remove deletion or make it harder to get to

----- Today December 16th, 2016 -----
benb [10:00 AM]  
Just to further complexify a bit in order to simplify later, what if we think about the 'knowledge garden' similar to how Permaculture design defines 5 zones for site development in increasing radius from the central living space

[10:01]  
So, user content can live in and move between say 3-5 zones of relevance and visibility based on a handful of factors, and behave accordingly

[10:03]  
Zone 1 is like active, primary visibility, while zone 3 and outwards is like topic limbo / archive / highly context specific use

[10:04]  
Just something to play with. Doesn't have to be strictly binary in/out, live/deleted etc

connor [12:38 PM]  
that seems resonant with thinking I’ve been doing on that matter as well ^^

Connoropolous:
there's also more background and discussion at issues:
#192 , #193 , #163 , #152 , and #63

Bortseb:
I feel like we should make it so that nothing is every truly deleted... to be able to undo something like a delete we need to have all the necessary metadata to be able rebuild the deleted entity with theUNDO.... so why not just keep that entity around.... and as long as it doesn't have any mappings it is effectively deleted...

Maybe we make "delete" mean something more like... this sthing wont show up in a search or auto-complete... but it is still technically in the system... waiting to be restored.

Connoropolous:
agreed Rob, that's the idea here

Bortseb:
Just to confirm... You can still choose to delete something, as long as you have all the right permissions... And this might actually affect other people's maps too.... But since we are enabling map change notifications, and UNDO.... if someone else wants to restore that topic to their map they can. Potentially just creating a duplicate of the same thing, but this time they are the creator of it (I realize UNDO isn't actually ready) Deleted topics wont show up in search or auto-complete...

Something very similar to delete is removing that entity so that it has no mappings... HOWEVER, this removed entity WILL show up in search and auto-complete... because its not deleted.

PR #999, opened 2016-12-20
[experimental] Realtime Streaming to Topic View

Connoropolous:


devvmh:
cool! so this is doing realtime on topic view, but going via action cable to the rails server instead of to the realtime server, right? very nice. i just skimmed it for now since it's in progress

Connoropolous:
not going to pursue this right now, expanded in a useful way on this work in #1010

PR #1010, opened 2016-12-29
Stream map related events from rails server via actioncable instead of nodejs

Connoropolous:
- [x] check security


app/channels/map_channel.rb
```diff
@@ -0,0 +1,8 @@
+class MapChannel < ApplicationCable::Channel
+  # Called when the consumer has successfully
+  # become a subscriber of this channel.
+  def subscribed
+    # TODO: verify permission
```
Connoropolous: I'm not precisely sure whether this is the appropriate place to determine whether the authenticated client has access to this stream of events related to a map. block people without permission to view the map from subscribing to events

config/initializers/warden_hooks.rb
```diff
@@ -0,0 +1,10 @@
+Warden::Manager.after_set_user do |user,auth,opts|
+  scope = opts[:scope]
+  auth.cookies.signed["#{scope}.id"] = user.id
+  auth.cookies.signed["#{scope}.expires_at"] = 30.minutes.from_now
+end
```
devvmh: after_set_user happens on each request like we were hoping

https://github.com/hassox/warden/wiki/Callbacks#after_set_user

PR #1019, opened 2017-01-09
tweak chat emoji

devvmh:


frontend/src/components/MapChat/NewMessage.js
```diff
@@ -41,7 +42,7 @@ class NewMessage extends Component {
           title="Emoji"
         />
         <div className="extra-message-options">
-          <span className="emoji-picker-button" onClick={this.toggleEmojiPicker}>🙃</span>
+          <span className="emoji-picker-button" onClick={this.toggleEmojiPicker}><Emoji size={24} emoji="upside_down_face" /></span>
```
Connoropolous: so there is a way to render emoji from the emoji mart?

devvmh: yes, the big issue is rendering it inside a textarea/message

textarea is much harder, but even inside a message div, how do you turn

"hello there 😛  more text 🇺🇸 and more"

into a renderable array

"hello there ",
<Emoji ... />,
" more text ",
<Emoji ... />,
" and more"

Connoropolous: right

Issue #845, opened 2016-10-26
Move screenshot functions out of the main map view


devvmh:
Here's the steps:

**v3.2** (done in #1027)
- [x] Add a button to download a screenshot to the import/export dialogue
- [x] Add a button to update the map screenshot to the map card
- [x] Add a deprecation notice when users use the old screenshot button (?) E.g. "This button is going away! The download feature is in the import/export dialogue above, and to change the map thumbnail check the map card in the (i) below!"

**v3.5**
- [ ] remove the old screenshot button

PR #1027, opened 2017-01-11
split screenshot function/button into two parts

devvmh:
See #845. Split the takeScreenshot button/function into two parts:

1. button in the import/export dialog to download a screenshot
2. button in the map card to update the map thumbnail

TODO:

- [x] add a warning when you use the old button on the map, noting the button on the map is going away
- [x] @connoropolous plz halp with the styling for the map card button
- [x] What very short label should be on the button in the map card? (see image below)
- [x] fix UX because of the multiple notifications when you click the old button

screen shot 2017-01-11 at 6 46 55 pm

### DIFF
--- a/app/assets/stylesheets/application.scss.erb
+++ b/app/assets/stylesheets/application.scss.erb
@@ -1558,6 +1558,7 @@ h3.filterBox {
     box-sizing: border-box;
     margin: 0.75em;
     padding: 0.75em;
+    padding-top: 0.85em;
     height: 3em;
     background-color: #AAB0FB;
     border-radius: 0.3em;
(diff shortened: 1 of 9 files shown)


Connoropolous:
Still need my help with the styling?

devvmh:
I made it look ok, but any fixes you want to do are welcome. You could also wait until I fix the UX issues with the GlobalUI.notifyUser calls

devvmh:
ready for review/testing. we could deploy to heroku. i'm anticipating we'll want to lower the timeout on some of the notifications... not totally sure exactly how it should work. clicking the deprecated button is quite the experience right now

PR #1031, opened 2017-01-13
topic card in react

Connoropolous:
Todo list:

- [x] make the riek desc textarea tall enough
- [x] disable markdown images in desc
- [x] break out attachments into their own component
- [x] move all listener binding to react or at least into the proper component
- [x] tiny pixel-wide line below metacode type when rolling over
- [x] fix metacode selection
- [x] close permission selector on selection
- [x] ensure text editing isn't possible by those without edit rights to the topic
- [x] fix draggability
- [x] fix embedly
- [x] convert metacode selection to react - i.e. no dangerouslySetInnerHTML
- [x] topic name riek editor also needs to be a text area so it can be multiline
- [x] fix scenario where user right clicks before opening topic card and #metacodeOptions is undefined
- [x] re-add topic title char counter
- [x] removing and adding an embedly link fails
- [x]  fix the remove link button to also clear any text in the link input field
- [x] pen icon is missing when i roll over title
- [x] fix the styling on the links
- [x] "Click to see which maps this topic appears on" doesn't seem to be working
- [x] clicking anywhere on the metacode title bar should open the selector
- [x] selecting a metacode should close the selector


frontend/src/Metamaps/Views/TopicCard.js
```diff
@@ -0,0 +1,61 @@
...
+  openTopicCard: null, // stores the topic that's currently open
+  init: function() {
+    // initialize topic card draggability and resizability
+    $('.showcard').draggable({
+      handle: '.metacodeImage',
+      stop: function() {
+        $(this).height('auto')
+      }
+    })
+  },
```
devvmh: will this remain draggable after a re-render? what's the html?

```
<div class="showcard">
  <div id="showcard">
  </div>
</div>
```

is that right? and react renders into the inner one?

what's draggable? the whole card?

frontend/src/Metamaps/Views/TopicCard.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+    ReactDOM.render(
+      React.createElement(ReactTopicCard, topicCardObj),
+      document.getElementById('showcard')
+    )
+  },
+  showCard: function(node, opts) {
+    var self = TopicCard
+    if (!opts) opts = {}
```
devvmh: es6 has default arguments so you can do `function(node, opts = {})`

frontend/src/Metamaps/Views/TopicCard.js
Connoropolous: yeah, but it probably doesn't need to have both divs... I'll scrap the inner one. I intend to implement dragging, but it might work a bit different, and not use this code. I am doing a little bit of simple design exploration alongside this work

frontend/src/Metamaps/Views/TopicCard.js
Connoropolous: right :)

frontend/src/Metamaps/Views/TopicCard.js
devvmh: Actually i was thinking 2 divs is fine

Connoropolous:
@devvmh if you do end up touching this, there is a chunk of work related to re-enabling click-to-edit using react instead of best-in-place. I don't think it's actually that hard a thing, except for maybe there's some nuances to the keyboard commands like making 'esc' and 'enter' submit the current changes, vs canceling. so that could be focused on, and don't touch the 'info bar' that divides the title and description, because I am going to be bringing it up for discussion really soon in terms of some possible changes. Green means go, red means stop :p

topic card highlighting

devvmh:
@Connoropolous embedly is sort of fixed, but if you remove a link to add a new one it fails to load until refresh. i'm a bit stumped. refreshing the page will render the correct new link though so it must be close

I considered a more native react solution like https://github.com/OsukeUesugi/react-embedly, except this project has 7 commits, the last commit was 9 months ago, and it's only got 4 stars. so i'm not super excited about adding that to metamaps...

not sure how to proceed with that but it's definitely the biggest issue blocking this PR. i'm gonna work on metacode selection before going back to embedliy

frontend/src/components/TopicCard/index.js
```diff
@@ -0,0 +1,55 @@
...
+      classname += ' cannotEdit'
+    }
+    if (topic.authorizePermissionChange(ActiveMapper)) classname += ' yourTopic'
+
+    return (
+      <div className={classname}>
+        <div className={`CardOnGraph ${hasAttachment ? 'hasAttachment' : ''}`} id={`topic_${topicId}`}>
+          <Title name={topic.get('name')}
+            onChange={this.props.updateTopic}
+            authorizedToEdit={topic.authorizeToEdit(ActiveMapper)}
```
Connoropolous: we've got authorizedToEdit defined like this at the top of the function

frontend/src/components/TopicCard/index.js
```diff
@@ -0,0 +1,55 @@
...
+          />
+          <Links topic={topic}
+            ActiveMapper={this.props.ActiveMapper}
+            updateTopic={this.props.updateTopic}
+          />
+          <Desc desc={topic.get('desc')}
+            authorizedToEdit={topic.authorizeToEdit(ActiveMapper)}
```
Connoropolous: we've got authorizedToEdit defined like this at the top of the function

Connoropolous:
yeah, the embedly bit is so tricky. that's why I delayed doing the topic card iin react as long as I did. Hmm, I will take another crack at it, if you're doing metacode selection stuff, and we'll see if we can close it off

Connoropolous:
this frickin thing :p what a trek

devvmh:
FYI the title char counter is proving to be REALLY hard because of the way riek is implemented. I'm considering a few options: 1) drop the title char counter 2) replace riek 3) patch or PR riek so it works 4) something even more clever that is the best of all worlds

devvmh:
@Connoropolous i'm checking off "fix the styling on the links" but there may be something I've missed. Can you make sure you review that before this gets pushed  tomake sure the styling is OK?

there are a few small style differences between this card and the current card, but i don't think anything is particularly bad that i've noticed. some of it i think is a bit nicer ;)

Connoropolous:
@devvmh I'm going to deploy another branch to heroku for a minute, but after I do that I'll switch it back to this branch, and check it out!

.agignore
```diff
@@ -0,0 +1 @@
+app/assets/javascripts/metamaps.secret.bundle.js
```
devvmh: i use the "silver searcher" to recursively search for strings in the codebase. I kept getting results in this file, so I ignored it. 

I could remove this file if metamaps.secret.bundle.js were in the "webpacked" directory, which is gitignored. But I guess the current approach has this file checked into git (which is probably bad). Ideally we would have this as a second bundle generated by webpack.config.js

frontend/src/Metamaps/Util.js
```diff
@@ -135,9 +135,26 @@ const Util = {
...
+        const imageSrc = node.destination
+        const textNode = new Node('text', node.sourcepos)
+        textNode.literal = `![${imageAlt}](${imageSrc})`
+
+        node.insertBefore(textNode)
+        node.unlink() // remove the image, replacing it with markdown
+        walker.resumeAt(textNode, false)
+      }
+    }
+
```
devvmh: this is not necessarily related to this PR. I probably shouldn't have put these together :(

Connoropolous:
so close! 
I think we should still have nil text for the description, if you can edit. I think it can just say 'edit description...' but it should look like placeholder text. 

also, something broked the right click metacode styling
screen shot 2017-02-25 at 7 11 29 pm

Connoropolous:
it's crazy to think of the code for all these new components being jumbled up in one file that uses jquery to manipulate all that, erggg

Connoropolous:
won't miss that!

devvmh:
@connoropolous i can't replicate the problem with the metacode selector css, it seems fine to me on localhost

devvmh:
nil text missing was a bug. I've fixed the bug so now it says 'Click here to add a description'

Connoropolous:
@devvmh done and done! thanks for trekking together on this :p

Issue #124, opened 2013-01-28
Enable file uploading in topic cards.


poietik:
the issue as framed for design: https://trello.com/c/iY26rzDE/58-uploading-images-files-to-topic-card

the issue as framed from the users perspective: https://trello.com/c/jmIs8gXb/44-uploading-images-files-to-topic-card

devvmh:
I should finish #1031 (rebuild topic card in react) before continuing on this

PR #1043, opened 2017-01-22
file attachments in db (re: #124)

devvmh:
this adds a new table `attachments`, allowing multiple files of different types to be attached to topics. In future, we could attach them to messages too, and then potentially other content via the polymorphic relationship.

Just for the backend, we have a small TODO list:

- [x] come up with a list of file types attachable to topics
- [x] make a module to be included in the Topic model that has functions for differentiating the different file types and accessing them smoothly (e.g. topic.attachments, topic.images, topic.documents, ...)
- [x] test the validations for the approved whitelist of types

Then in the future we'll have a few more broad issue to do. I'd be OK merging this PR before doing this list:

- [ ] UI for adding files to topics
- [ ] UI for displaying and accessing attached files on topics
- [ ] add file attachments to messages
- [ ] UI for uploading files in chat to messages
- [ ] UI for displaying files in chat messages


devvmh:
I ran `rubocop -a` on the config directory, so there's some non-important rubocop style fixes mixed in with this PR on the config directory

PR #1045, opened 2017-01-23
browser-support.md

devvmh:
looking at https://kangax.github.io/compat-table/es5/, i think it’s reasonable to support

IE9+
Edge 12+
FF 38+
Chrome 46+
Opera 33+
Safari 6+
Android 4.4+
iOS 6+

that previous list is aimed at browsers that mostly support es5, which is a bare minimum i think. if we wanted to push it to es6 support, we could say

no IE
Edge 13+
FF 45+ (latest ESR release)
Chrome 55+ (latest stable release)
Safari 10+
iOS 10+
Android - only support chrome

I don’t think this second list is as reasonable.

as a reference, basecamp’s support is a little harsher than the es5 list above: https://basecamp.com/help/3/guides/account/browsers


Connoropolous:
If we are claiming to support all these, then I think there'd be some work to do to actually back up that claim. I think its best on chrome, and I have a suspicion there's a variety of flaws that might show themselves if we went through the app in depth on other browsers. even if its just styling.

devvmh:
yeah, which is a good point. i've been trying to limit my enthusiasm for new js features so that we continue to support legacyish browsers. 

I do think it's worth it to have broad support similar to this, especially if we want to support large organizations with older computers.

We could *aim* to support these, and not *claim* to support them. Not sure what they best way to do that is; maybe just adding a note to the top of this file

We could also be dicks and just support the es6 list but i don't think it's a good strategy

maybe we should put out a call for testers with old browsers on twitter

there are also tools to test compatibility, but they won't be very comprehensive

PR #1048, opened 2017-01-26
track pageloads with google analytics. fixes #1012

### LINKED ISSUE #1012: track pageloads with GA better
add `ga('send', 'pageview');` to the code every time we use the backbonejs history pushState function, so that google analytics can track non-page-reload page changes

Connoropolous:


Connoropolous:
and @devvmh as long as we like this one, this would be a good one to hotfix in to metamaps.cc so we can have more insight in the analytics sooner rather than later

frontend/src/Metamaps/Router.js
```diff
@@ -202,6 +202,11 @@ const _Router = Backbone.Router.extend({
 
 const Router = new _Router()
 
+Router.navigateAndTrack = (fragment, options) => {
+  Router.navigate(fragment, options)
+  ga && ga('send', 'pageview')
```
devvmh: Can you change this to `window.ga && window.ga('send', 'pageview')`? Stuff like this always causes trouble if you're in a node env (e.g. testing)

devvmh:
LGTM other than that one comment. I've just now hotfixed this into metamaps.cc

frontend/src/Metamaps/Router.js
Connoropolous: Yep

PR #1054, opened 2017-01-28
searchfields api parameter

devvmh:
I'm open to suggestions for better names for this parameter...


Connoropolous:
this is great, it will work. I guess in the future, we will surely be working to define a more fully baked querying solution

PR #1059, opened 2017-02-05
Secret.convo

Connoropolous:
- [ ] still to maybe do, block this route from users who aren't "one of us"?

everything about this PR is be entirely non-user-impacting, besides those who can access this special feature

`metamaps.secret.bundle` is compiled from the code found on the `feature/convo.algo` branch


app/policies/map_policy.rb
```diff
@@ -20,6 +20,10 @@ def show?
       record.collaborators.include?(user) ||
       record.user == user
   end
+  
+  def conversation?
+    show?
+  end
```
devvmh: you could use this to filter on user email if you want

app/assets/javascripts/application-secret.js
```diff
@@ -0,0 +1,23 @@
...
+//= require jquery
+//= require jquery-ui
+//= require jquery_ujs
+//= require action_cable
+//= require_directory ./lib
+//= require ./cloudcarousel-secret
+//= require ./metamaps.secret.bundle
+//= require ./Metamaps.ServerData
+//= require homepageVimeoFallback
+/* eslint-enable spaced-comment */
```
devvmh: Why are you using multiple bundles exactly? I'm not opposed, just interested

Maybe we could use this as an opportunity to implement a real AB testing framework. I've started researching this here: https://www.reddit.com/r/javascript/comments/3x2b56/ab_testing_and_single_page_application_good/. I was considering just using code splitting, but now I'm thinking about how we could set that up to be integrated like this with different bundles.

devvmh:
You'll need to update the rspec tests so travis will pass. I guess we just remove the tests verifying that xloc/yloc are present!

app/assets/javascripts/application-secret.js
Connoropolous: "Why are you using multiple bundles exactly? I'm not opposed, just interested"

Because we've been ripping and tearing at all the J's code on the other branch to make this special mode work, without trying to keep the code workable for the main mode at the same time. Once we are on to the path we really like we do it better, and make the two modes able to co-exist in the same bundle/codebase

app/policies/map_policy.rb
Connoropolous: Yeah, I think I will

Connoropolous:
K, will fix the tests then merge. We'll give it a run on heroku, but I'm hoping to get it up on metamaps asap, we are trying to do some stuff with it throughout February

app/assets/javascripts/application-secret.js
Connoropolous: P.s. @devvmh I'm interested in exploring this AB testing pathway too!

PR #1063, opened 2017-02-08
Follows and some new notifications

Connoropolous:
- [ ] figure out how to handle permissions in notifications

- [x] figure out how to handle situations where dual notifications would get sent, due to the person following both topics connected by a synapse, or a topic & the map its gets added to
-- return a list of the people it got sent to, and pass in those users as an exclude for the other notification


app/models/synapse.rb
```diff
@@ -77,10 +85,18 @@ def after_updated
       old = changed_attributes.select { |k| attrs.include?(k) }
       meta = new.merge(old) # we are prioritizing the old values, keeping them
       meta['changed'] = changed_attributes.keys.select { |k| attrs.include?(k) }
+      
+      # FIXME: there's a bug here, the topic creator is not always the updater
```
Connoropolous: TODO: fix this with this PR

app/models/topic.rb
```diff
@@ -155,10 +164,20 @@ def after_updated
       old = changed_attributes.select { |k| attrs.include?(k) }
       meta = new.merge(old) # we are prioritizing the old values, keeping them
       meta['changed'] = changed_attributes.keys.select { |k| attrs.include?(k) }
+      
+      # FIXME: there's a bug here, the topic creator is not always the updater
```
Connoropolous: TODO: fix this too in this PR

app/services/follow_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+class FollowService
+  
+
+  def self.follow(entity, user, reason)
+
+    #return unless is_tester(user)
```
Connoropolous: Uncomment this if we push to prod

Connoropolous:
@devvmh I requested a review even though this isn't ready ready yet. If you don't have bandwidth for it right now with the job starting no worries :)

Connoropolous:
screen shot 2017-02-08 at 2 17 23 am

app/controllers/synapses_controller.rb
```diff
@@ -47,9 +49,11 @@ def update
     @synapse = Synapse.find(params[:id])
     @synapse.desc = '' if @synapse.desc.nil?
     authorize @synapse
+    @synapse.updated_by = current_user
+    @synapse.assign_attributes(synapse_params)
```
devvmh: are these two lines in the wrong order?

app/mailers/application_mailer.rb
```diff
@@ -9,15 +9,20 @@ def deliver
...
+        when MAILBOXER_CODE_ACCESS_APPROVED
+          request = notification.notified_object
+          MapMailer.access_approved(request)
+        when MAILBOXER_CODE_INVITE_TO_EDIT
+          user_map = notification.notified_object
+          MapMailer.invite_to_edit(user_map)
+        when MAILBOXER_CODE_MAP_MESSAGE
+        when MAILBOXER_CODE_MAP_STARRED
+        when MAILBOXER_CODE_TOPIC_ADDED_TO_MAP
+        when MAILBOXER_CODE_TOPIC_CONNECTED
```
devvmh: this is great. one thought i have about this is that this approach, where it's built into application_mailer, might be too limiting. I was naively hoping notifications would remain simple, but like 2 days after i merged mailboxer it was clear they wouldn't ahaha. i should have guessed that anyways...

but yeah we might need to change this approach later. just noting that.

app/models/access_request.rb
```diff
@@ -25,12 +27,11 @@ def deny
...
-
-  def approved_text
-    map.name + ' - access approved'
+  
+  protected
+  
+  def after_created_async
+    NotificationService.access_request(self)
   end
+  handle_asynchronously :after_created_async
```
devvmh: oh now i get this code. awesome :+1:

app/services/notification_service.rb
```diff
@@ -7,32 +7,163 @@ def self.renderer
...
+      when 'topic_connected'
+        MAILBOXER_CODE_TOPIC_CONNECTED
+      when 'topic_deleted'
+        MAILBOXER_CODE_TOPIC_DELETED
+      when 'topic_disconnected'
+        MAILBOXER_CODE_TOPIC_DISCONNECTED
+      when 'topic_updated'
+        MAILBOXER_CODE_TOPIC_UPDATED
+    end
+  end
```
devvmh: I'm not opposed to using these event names in mailboxer. but i also see the argument for leaving it as is, in case (or when) the events and the notifications begin to diverge more and more

app/services/notification_service.rb
```diff
@@ -7,32 +7,163 @@ def self.renderer
...
+      when 'topic_connected' # event is a Synapse
+        'new synapse to topic ' + entity.name
+      when 'topic_deleted' #disabled
+        'placeholder'
+      when 'topic_disconnected' #disabled
+        'placeholder'
+      when 'topic_updated' #disabled
+        'placeholder'
+    end
+  end
```
devvmh: this might be overkill. can't this go in individual mailers? if not, whateva, but it seems too centralized to get the subject here

spec/factories/topics.rb
```diff
@@ -2,6 +2,7 @@
 FactoryGirl.define do
   factory :topic do
     user
+    association :updated_by, factory: :user
```
devvmh: these updated_by fields should be checked somewhere, right? i don't think that's currently happening

app/controllers/synapses_controller.rb
Connoropolous: makes no diff either way I don't believe

app/mailers/application_mailer.rb
Connoropolous: yeah...

app/services/notification_service.rb
Connoropolous: what is "as is"?

spec/factories/topics.rb
Connoropolous: checked how do you mean?

Connoropolous: tested? I was getting lost trying to test them

app/models/access_request.rb
Connoropolous: yeah, i was implementing a new standard everywhere else, so I wanted to modify this case to match that

app/controllers/synapses_controller.rb
devvmh: ic, as long as updated_by is never in synapse_params

app/services/notification_service.rb
devvmh: i mean we could delete all these constants like MAILBOXER_CODE_TOPIC_UPDATED and just use `'topic_updated'`

spec/factories/topics.rb
devvmh: yes tested

app/controllers/synapses_controller.rb
Connoropolous: which it should be, because then you could mock as if someone else made the change

Connoropolous: shouldn't be*

app/services/notification_service.rb
Connoropolous: i c. yeah, I may do that. the rest is kind of extraneous

Connoropolous:
yus, I did it. @devvmh 

that was hard. but worth it

Connoropolous:
this prs changes are substantially smaller now

app/services/notification_service.rb
Connoropolous: True

PR #1064, opened 2017-02-09
helper function and notification setup for messages from the devs to be sent to all users

devvmh:


app/services/notification_service.rb
```diff
@@ -23,16 +26,29 @@ def self.invite_to_edit(map, inviter, invited)
...
+
   def self.text_for_notification(notification)
     if notification.notification_code == MAILBOXER_CODE_ACCESS_REQUEST
-      map = notification.notified_object.map
+      map = notification.notified_object&.map
       'wants permission to map with you on <span class="in-bold">' + map.name + '</span>&nbsp;&nbsp;<div class="action">Offer a response</div>'
     elsif notification.notification_code == MAILBOXER_CODE_ACCESS_APPROVED
-      map = notification.notified_object.map
+      map = notification.notified_object&.map
       'granted your request to edit map <span class="in-bold">' + map.name + '</span>'
```
Connoropolous: This will still throw an error, because of the lines that follow the one that you fixed. The best thing to do might be to filter out notifications in the first place, that don't have a non-nil "notified_object"

Connoropolous: I'll fix this

PR #1071, opened 2017-02-12
fix travis

devvmh:


app/services/follow_service.rb
```diff
@@ -1,35 +1,35 @@
...
-  protected
-  
-  def is_tester(user)
-    %w(connorturland@gmail.com devin@callysto.com chessscholar@gmail.com solaureum@gmail.com ishanshapiro@gmail.com).include?(user.email)
+    
+    protected
+    
+    def is_tester(user)
+      %w(connorturland@gmail.com devin@callysto.com chessscholar@gmail.com solaureum@gmail.com ishanshapiro@gmail.com).include?(user.email)
+    end
```
devvmh: @Connoropolous  can you review this class? or just merge it and make sure the functionality works correctly haha.

I originally modified line 31 to say `def self.is_tester(user)`. But then I just wrapped the whole class in `class << self` because it seems fancier

PR #1081, opened 2017-03-05
daily map activity emails

Connoropolous:
got those tests passing @devvmh !


app/models/mapping.rb
```diff
@@ -41,10 +41,11 @@ def after_created
         topic2: mappable.topic2.filtered,
         mapping_id: id
       )
-      Events::SynapseAddedToMap.publish!(mappable, map, user, nil)
+      meta = { 'mapping_id': id }
+      Events::SynapseAddedToMap.publish!(mappable, map, user, meta)
```
devvmh: was this a bug or something?

app/services/map_activity_service.rb
```diff
@@ -0,0 +1,90 @@
+class MapActivityService
+
+  def self.summarize_data(map, user, till = DateTime.now)
+    results = {
+      stats: {}
+    }
+
+    since = till - 24.hours
```
devvmh: can we call this until instead of till 😂

app/services/map_activity_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    topics_removed_events = Event.where(kind: 'topic_removed_from_map', map: map)
+                                 .where("created_at > ? AND created_at < ?", since, till)
+                                 .where.not(user: user)
+                                 .order(:created_at)
+
+    topics_added_to_include = {}
+    topics_added_events.each do |ta|
+      num_adds = topics_added_events.where(eventable_id: ta.eventable_id).count
+      num_removes = topics_removed_events.where(eventable_id: ta.eventable_id).count
+      topics_added_to_include[ta.eventable_id] = ta if num_adds > num_removes
```
devvmh: nice

lib/tasks/emails.rake
```diff
@@ -0,0 +1,20 @@
...
+    summarize_map_activity
+  end
+
+  def summarize_map_activity
+    Map.all.find_each do |map|
+      map.followers.each do |user|
+        # add logging and rescue-ing
+        # and a notification of failure
+        next unless MapPolicy.new(user, map).show? # just in case the permission changed
+        next unless user.emails_allowed
```
devvmh: nice, these two lines are v clever

are you going to update the email permissioning in this PR? It seems like the right time to do it. I could maybe push some commits to the branch if it helps

spec/factories/message.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+FactoryGirl.define do
+  factory :message do
+    association :resource, factory: :map
+    user
+    sequence(:message) { |n| "Cool Message ##{n}" }
```
devvmh: yesssss Cool Message good choice

spec/services/map_activity_service.rb
```diff
@@ -0,0 +1,286 @@
+require 'rails_helper'
+
+RSpec.describe MapActivityService do
+  let(:map) { create(:map, created_at: 1.week.ago) }
+  let(:other_user) { create(:user) }
+  let(:email_user) { create(:user) }
+
+  it 'includes nothing if nothing happened' do
+    response = MapActivityService.summarize_data(map, email_user)
+    expect(response[:stats]).to eq({})
```
devvmh: i'd argue we should be testing the whole hash to be equal to { stats: {} }

We could even make that a constant to check against during these tests

It just communicates more info

devvmh:
sweeeet!! this looks great

app/models/mapping.rb
Connoropolous: Yeah, so to speak. I hadn't thought of any reason prior to this why it would have been necessary

app/services/map_activity_service.rb
Connoropolous: Yee

lib/tasks/emails.rake
Connoropolous: Which email permissions? For the other notifications?

spec/services/map_activity_service.rb
Connoropolous: Cool, yeah

lib/tasks/emails.rake
devvmh: I guess i mean email preferences

Connoropolous: To make it more granular? Yeah, now would be a good time. I have ideas abt it I'll work on something

lib/tasks/emails.rake
```diff
@@ -0,0 +1,20 @@
+namespace :metamaps do
+  desc "delivers recent map activity digest emails to users"
+  task deliver_map_activity_emails: :environment do
+    summarize_map_activity
+  end
+
+  def summarize_map_activity
+    Follow.where(followed_type: 'Map').find_each do |follow|
```
Connoropolous: @devvmh I realized we can constrain the process to only call what needs to be called by looking for follows instead of for maps. I guess I might want to eager load the maps and users associations

app/services/map_activity_service.rb
```diff
@@ -0,0 +1,94 @@
...
+      num_adds = topics_added_events.where(eventable_id: ta.eventable_id).count
+      num_removes = topics_removed_events.where(eventable_id: ta.eventable_id).count
+      topics_removed_to_include[ta.eventable_id] = ta if num_removes > num_adds
+    end
+    if topics_removed_to_include.keys.length > 0
+      results[:stats][:topics_removed] = topics_removed_to_include.keys.length
+      results[:topics_removed] = topics_removed_to_include.values
+    end
+
+    synapses_added_events = Event.where(kind: 'synapse_added_to_map', map: map)
```
Connoropolous: Realized I need to get down to granular permissioning here, related to topics and synapses because stuff can be going on secretly on a map and other people shouldn't know.

spec/services/map_activity_service.rb
```diff
@@ -0,0 +1,287 @@
...
+      response = MapActivityService.summarize_data(map, email_user)
+      expect(response[:stats][:messages_sent]).to eq(1)
+    end
+
+    it 'does not count messages sent by the person who will receive the data' do
+      create(:message, resource: map, created_at: 5.hours.ago, user: other_user)
+      create(:message, resource: map, created_at: 5.hours.ago, user: email_user)
+      response = MapActivityService.summarize_data(map, email_user)
+      expect(response[:stats][:messages_sent]).to eq(1)
+    end
```
Connoropolous: add tests that check permissions of topics and synapses

Connoropolous:
As I mentioned to Devin: 

I had a thought for the email activity thing... we should be able to pass in the 'since' value I think, and then we could store the value that the most recent email was sent at on the Follow. That way, we would always ensure that each email summarized activity since the last summary was sent. Which should be approximately 24 hours, but as we've discussed that's not guaranteeed  
and if there is no 'since' value stored on the follow, default to 24 hours ago
it would give us an option for rerunning the script too, if it fails

and he agreed. but I'm going to merge this branch in anyways because its fine as is, and we can get these running, and then implementation of design can happen over the next little bit. Because this is all based on 'following' and we are the only people who can currently follow anything, only we will begin to receive these emails for time being

PR #1083, opened 2017-03-06
Fixes bug where pressing delete key while editing text will suggest...

Bortseb:
the deletion of selected map entities


Bortseb:
actually hold off on looking at this.. I forgot you requested that things got removed instead

PR #1103, opened 2017-03-23
[very wip] refactoring so that data visualizations won't conflict, by making them scoped not global

Connoropolous:


Connoropolous:
abandoning this particular effort. not worth it

PR #1106, opened 2017-03-26
update vimeo url, use asset path + erb for poster path

devvmh:


devvmh:
This is hotfixed on metamaps.cc

PR #1109, opened 2017-04-10
Improvements to Import

Bortseb:
Changed URL Regex to make more links importable, also removed need for special text formatting in order to paste or drop new topic labels. (Didn't break TSV import mode)

### DIFF
--- a/frontend/src/Metamaps/Import.js
+++ b/frontend/src/Metamaps/Import.js
@@ -388,6 +388,36 @@ const Import = {
       }
     )
   },
+  
+  handleTEXT: function(text, opts = {}) {
+    let coords = opts.coords
+    if (!coords || coords.x === undefined || coords.y === undefined) {
+      coords = AutoLayout.getNextCoord({ mappings: DataModel.Mappings })
+    }
+
+    const name = text
+    const url = ""
+    const metacode = opts.metacode || 'Wildcard'
+    const importId = opts.importId || null // don't store a cidMapping
+    const permission = opts.permission || null // use default
+    const desc = opts.desc || ""
+
+    Import.createTopicWithParameters(
+      name,
+      metacode,
+      permission,
+      desc,
+      url,
+      coords.x,
+      coords.y,
+      importId,
+      {
+        success: function(topic) {
+          return
+        }
+      }
+    )
+  },
 
   /*
    * helper functions
--- a/frontend/src/Metamaps/Import.js
+++ b/frontend/src/Metamaps/Import.js
@@ -388,7 +388,7 @@ const Import = {
     )
   },
 
-  handleTEXT: function(text, opts = {}) {
+  handleText: function(text, opts = {}) {
     let coords = opts.coords
     if (!coords || coords.x === undefined || coords.y === undefined) {
       coords = AutoLayout.getNextCoord({ mappings: DataModel.Mappings })
@@ -409,12 +409,7 @@ const Import = {
       url,
       coords.x,
       coords.y,
-      importId,
-      {
-        success: function(topic) {
-          return
-        }
-      }
+      importId
     )
   },
 
--- a/frontend/src/Metamaps/Import.js
+++ b/frontend/src/Metamaps/Import.js
@@ -6,7 +6,6 @@ import _ from 'lodash'
 import Active from './Active'
 import AutoLayout from './AutoLayout'
 import DataModel from './DataModel'
-import GlobalUI from './GlobalUI'
 import Map from './Map'
 import Synapse from './Synapse'
 import Topic from './Topic'
@@ -388,7 +387,7 @@ const Import = {
       }
     )
   },
-  
+
   handleTEXT: function(text, opts = {}) {
     let coords = opts.coords
     if (!coords || coords.x === undefined || coords.y === undefined) {
@@ -396,11 +395,11 @@ const Import = {
     }
 
     const name = text
-    const url = ""
+    const url = ''
     const metacode = opts.metacode || 'Wildcard'
     const importId = opts.importId || null // don't store a cidMapping
     const permission = opts.permission || null // use default
-    const desc = opts.desc || ""
+    const desc = opts.desc || ''
 
     Import.createTopicWithParameters(
       name,
--- a/frontend/src/Metamaps/PasteInput.js
+++ b/frontend/src/Metamaps/PasteInput.js
@@ -68,11 +68,9 @@ const PasteInput = {
     } else if (text[0] === '{') {
       Import.handleJSON(text)
     } else if (text.match(/[Tt]opics\t/) || text.match(/[Ss]ynapses\t/)) {
-      console.log("Tab sep!")
       Import.handleTSV(text)
     } else {
       //Handle as plain text
-      console.log("The text is:",text,"!")
       let textItems = text.split("\n")
       if (textItems.length === 1){
         if(textItems[0].trim() !== ""){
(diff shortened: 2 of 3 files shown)


frontend/src/Metamaps/PasteInput.js
```diff
@@ -70,11 +67,23 @@ const PasteInput = {
       Import.handleURL(text, coords)
     } else if (text[0] === '{') {
       Import.handleJSON(text)
-    } else if (text.match(/\t/)) {
+    } else if (text.match(/[Tt]opics\t/) || text.match(/[Ss]ynapses\t/)) {
```
devvmh: I'd change these regexes to match the start of the string. I'd have to look it up, but I think it's `^` to match the start of a string

frontend/src/Metamaps/PasteInput.js
```diff
@@ -29,12 +28,10 @@ const PasteInput = {
       if (e.dataTransfer.files.length > 0) {
         self.handleFile(e.dataTransfer.files[0], coords)
       }
-      // OMG import bookmarks 😍
+      // OMG import bookmarks 😍 (Or just text :P)
```
devvmh: Correct me if I'm wrong, but I think this particular code path only happens for bookmarks. If you verify that by putting a debugger here and triggering it without a bookmark I'll believe you, but I think paste and drag-drop trigger at different points. This idea of a "dataTransfer.items" I think only happens for bookmarks.

frontend/src/Metamaps/PasteInput.js
```diff
@@ -7,8 +7,7 @@ import Visualize from './Visualize'
 const PasteInput = {
   // thanks to https://github.com/kevva/url-regex
   // eslint-disable-next-line no-useless-escape
-  URL_REGEX: new RegExp('^(?:(?:(?:[a-z]+:)?//)|www\.)(?:\S+(?::\S*)?@)?(?:localhost|(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])(?:\.(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])){3}|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,}))\.?)(?::\d{2,5})?(?:[/?#][^\s"]*)?$'),
-
+  URL_REGEX: new RegExp(/^(?:(?:https?|ftp):\/\/)(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,}))\.?)(?::\d{2,5})?(?:[/?#]\S*)?$/,"i"),
```
devvmh: Where's this url regex from? What does it improve? I believe it improves it I'm just interested + want to document it somehow

frontend/src/Metamaps/Import.js
```diff
@@ -388,6 +388,36 @@ const Import = {
       }
     )
   },
+  
+  handleTEXT: function(text, opts = {}) {
```
devvmh: This should be handleText to match the style in the rest of the repo

frontend/src/Metamaps/Import.js
```diff
@@ -388,6 +388,36 @@ const Import = {
...
+      metacode,
+      permission,
+      desc,
+      url,
+      coords.x,
+      coords.y,
+      importId,
+      {
+        success: function(topic) {
+          return
```
devvmh: You can just omit the success callback if it's not going to do anything (right? I'm pretty sure this function will allow it to be undefined)

frontend/src/Metamaps/PasteInput.js
```diff
@@ -70,11 +67,23 @@ const PasteInput = {
...
+      let textItems = text.split("\n")
+      if (textItems.length === 1){
+        if(textItems[0].trim() !== ""){
+            Import.handleTEXT(textItems[0].trim(),coords)
+        }
+      }
+      else if (window.confirm('Are you sure you want to create ' + textItems.length + ' new topics?')) {
+        textItems.forEach(item => {
+          if(item.trim() !== ""){
+            self.handle(item.trim(),coords)
```
devvmh: Can you run eslint and/or check out https://codeclimate.com/github/metamaps/metamaps/pull/1109 to fix simple style stuff.

frontend/src/Metamaps/PasteInput.js
devvmh: hmm i also see this disallows ip addresses like 192.168.x.x and 172.x.x.x or 169.254. I'm wondering if that's desirable. Probably it is.... maybe not if we make a corporate instance in the future and they want to have links to stuff on their internal network? It's an interesting question. Maybe I'm just prematurely optimizing

frontend/src/Metamaps/PasteInput.js
Bortseb: But you can drag and drop just raw text from other places too, that's what I meant with this... It doesn't have to be bookmarks from my testing... (Unless i'm just confused and everything I drag and drop is actually just simulating a paste event, and it is only dragged bookmarks/links that truly drop events

devvmh: I'm pretty sure it's a paste event, but I'd have to verify with console.log or something to be 100% sure. Are you able to verify that and update the comment? We could even add a comment in the correct place saying "drag and drop text and it'll do this"

frontend/src/Metamaps/PasteInput.js
devvmh: oh that is beautiful. i'll push an update to the branch that imports that regex from a file with the full documentation

frontend/src/Metamaps/PasteInput.js
devvmh: i checked it - you are right, if i drag text e.g. from the input at the top then it imports as an "item"

frontend/src/Metamaps/PasteInput.js
devvmh: I made this change on the branch - can you test it when you're back developing and make sure it's not screwing anything up

frontend/src/Metamaps/Import.js
devvmh: ah I get it now, every other function is like handleCSV handleTSV. I think since it's not an acronym i'll make this change on the branch so it's handleText

frontend/src/Metamaps/Import.js
devvmh: did this on the branch

devvmh:
OK I've addressed all the comments I had on this branch, so once you've given it another test to make sure I didn't screw anything up go ahead and merge 😁

devvmh:
nah, that's the JS version, the PHP version has underscores around it and you can't say "new RegExp" in PHP.

Bortseb:
Ah! I got confused because some of the regex changed when you compare the all-on-one-line version versus the commented multiline version... Anyhow, it works! :) SQUASH!

PR #1117, opened 2017-05-11
alert user if map deletion fails because it's been forked

devvmh:


devvmh:
close in favour of #1119

PR #1128, opened 2017-09-02
dont send notification for each topic of forked map

### LINKED ISSUE #1118: Topic notification Edge case: don't send notification for every topic when a map gets forked
Instead, send one notification saying the map was forked

Connoropolous:
fixes #1118

### DIFF
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -38,6 +38,8 @@ class NotificationService
         when TOPIC_ADDED_TO_MAP
           next unless TopicPolicy.new(follow.user, entity).show? && MapPolicy.new(follow.user, event.map).show?
           next if follow.user.has_map_open(event.map)
+          # note, the caveat to this fork related code is that no future updates will be sent
+          # either. only the notifications for the initial fork should be squashed like this
           next if event.map.source_id # if map is a fork, send notification for that instead
         when TOPIC_CONNECTED_1, TOPIC_CONNECTED_2
           next unless SynapsePolicy.new(follow.user, event).show?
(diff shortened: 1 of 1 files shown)


app/services/notification_service.rb
```diff
@@ -38,6 +38,7 @@ def self.send_for_follows(follows, entity, event_type, event)
         when TOPIC_ADDED_TO_MAP
           next unless TopicPolicy.new(follow.user, entity).show? && MapPolicy.new(follow.user, event.map).show?
           next if follow.user.has_map_open(event.map)
+          next if event.map.source_id # if map is a fork, send notification for that instead
         when TOPIC_CONNECTED_1, TOPIC_CONNECTED_2
```
devvmh: LGTM. We could consider an is_fork? function in thr future but this seems ok

Connoropolous: Hmm. Just realized that even after the initial forking action, notifications related to topics being added to this map will be squashed due to this line of code. I'll make a note of it in the code and go ahead with it.

app/services/notification_service.rb
```diff
@@ -38,6 +38,9 @@ def self.send_for_follows(follows, entity, event_type, event)
         when TOPIC_ADDED_TO_MAP
           next unless TopicPolicy.new(follow.user, entity).show? && MapPolicy.new(follow.user, event.map).show?
           next if follow.user.has_map_open(event.map)
+          # note, the caveat to this fork related code is that no future updates will be sent
+          # either. only the notifications for the initial fork should be squashed like this
+          next if event.map.source_id # if map is a fork, send notification for that instead
```
devvmh: I guess the options are:

1. store a field on a map that says "I'm a fork, and I have/have not already copied all the topics from my parent map". This could be unstable if there are errors during the forking though.
2. something else more complicated as part of a larger fork logic revamp
3. just check the created_at time here and don't send notifications for the first 30 seconds (this is suuuuper hacky and could lead to stuff being missed right after an initial fork, but since no one should be following the fork at first anyways, maybe it's ok??)

(2) seems best haha, although (3) is intriguing

Connoropolous: lol, I was thinking about doing number 3, but then opted not to, in the hopes that we'll be incentivized to do something more like 2 sooner :)

PR #1130, opened 2017-09-03
make notifs and follows work for all users

Connoropolous:
the change to the sign in line is just due to this deprecation warning message I saw.  I tested to make sure it still works and it does.
screen shot 2017-09-03 at 9 08 04 am

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -23,15 +23,15 @@ class UsersController < ApplicationController
     if user_params[:password] == '' && user_params[:password_confirmation] == ''
       # not trying to change the password
       if @user.update_attributes(user_params.except(:password, :password_confirmation))
-        update_follow_settings(@user, params[:settings]) if is_tester(@user)
+        update_follow_settings(@user, params[:settings])
         @user.image = nil if params[:remove_image] == '1'
         @user.save
-        sign_in(@user, bypass: true)
+        bypass_sign_in(@user)
         respond_to do |format|
           format.html { redirect_to root_url, notice: 'Settings updated' }
         end
       else
-        sign_in(@user, bypass: true)
+        bypass_sign_in(@user)
         respond_to do |format|
           format.html { redirect_to edit_user_path(@user), notice: @user.errors.to_a[0] }
         end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -58,8 +58,8 @@ class User < ApplicationRecord
       admin: admin }
     if (_options[:follows])
       json['follows'] = {
-        topics: following.where(followed_type: 'Topic').to_a.map(&:followed_id),
-        maps: following.where(followed_type: 'Map').to_a.map(&:followed_id)
+        topics: following.where(muted: false, followed_type: 'Topic').to_a.map(&:followed_id),
+        maps: following.where(muted: false, followed_type: 'Map').to_a.map(&:followed_id)
       }
     end
     if (_options[:follow_settings])
--- a/app/services/follow_service.rb
+++ b/app/services/follow_service.rb
@@ -8,13 +8,15 @@ class FollowService
       return if (reason == 'created' || reason == 'contributed') && !should_auto_follow(entity, user, reason)
 
       follow = Follow.where(followed: entity, user: user).first_or_create
+      follow.update_attribute('muted', false)
       if FollowReason::REASONS.include?(reason) && !follow.follow_reason.read_attribute(reason)
         follow.follow_reason.update_attribute(reason, true)
       end
     end
 
     def unfollow(entity, user)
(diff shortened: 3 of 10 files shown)


frontend/src/components/TopicCard/index.js
```diff
@@ -53,7 +52,7 @@ class ReactTopicCard extends Component {
                 authorizedToEdit={authorizedToEdit}
                 updateTopic={wrappedUpdateTopic}
               />
-            {Util.isTester(currentUser) && <Follow isFollowing={isFollowing} onTopicFollow={wrappedTopicFollow} />}
+            <Follow isFollowing={isFollowing} onTopicFollow={wrappedTopicFollow} />
```
devvmh: is the styling on this still a bit ugly? can we upgrade a bit as part of this? even just some minimal changes

app/controllers/users_controller.rb
```diff
@@ -23,15 +23,15 @@ def update
...
         @user.image = nil if params[:remove_image] == '1'
         @user.save
-        sign_in(@user, bypass: true)
+        bypass_sign_in(@user)
         respond_to do |format|
           format.html { redirect_to root_url, notice: 'Settings updated' }
         end
       else
-        sign_in(@user, bypass: true)
+        bypass_sign_in(@user)
```
devvmh: were you able to test this works the same way (and works for us especially)?

Connoropolous: yep!

frontend/src/components/TopicCard/index.js
Connoropolous: It's not ugly, but it's also not pretty. I'll see if Pegah can just help me improve how it looks and works a little bit

app/services/follow_service.rb
```diff
@@ -31,6 +33,8 @@ def remove_reason(entity, user, reason)
     protected
 
     def should_auto_follow(entity, user, reason)
+      follow = Follow.where(followed: entity, user: user).first
+      return false if follow && follow.muted
```
Connoropolous: This prevents the auto-follow from happening, when you've manually unfollowed it

app/services/notification_service.rb
```diff
@@ -51,7 +51,8 @@ def self.send_for_follows(follows, entity, event_type, event)
 
   def self.notify_followers(entity, event_type, event, reason_filter = nil, exclude_follows = nil)
     follows = entity.follows.where.not(user_id: event.user.id)
-    
+    follows = follows.where(muted: false)
```
Connoropolous: this prevents notifications from going out to follows marked as 'muted'

app/services/follow_service.rb
```diff
@@ -3,18 +3,20 @@ class FollowService
   class << self 
     def follow(entity, user, reason)
 
-      return unless user && is_tester(user)
+      return unless user
 
       return if (reason == 'created' || reason == 'contributed') && !should_auto_follow(entity, user, reason)
 
       follow = Follow.where(followed: entity, user: user).first_or_create
+      follow.update_attribute('muted', false)
```
Connoropolous: This ensures a muted follow flips back to normal when you manually follow the entity

app/models/user.rb
```diff
@@ -58,8 +58,8 @@ def as_json(_options = {})
       admin: admin }
     if (_options[:follows])
       json['follows'] = {
-        topics: following.where(followed_type: 'Topic').to_a.map(&:followed_id),
-        maps: following.where(followed_type: 'Map').to_a.map(&:followed_id)
+        topics: following.where(muted: false, followed_type: 'Topic').to_a.map(&:followed_id),
```
Connoropolous: These just treat muted follows as if they don't exist. There's one tiny caveat here, which is that the front end doesn't know in which circumstances to avoid flipping a non-follow to a follow, in the case of an edit, but I am ok with that tiny gap, which will be resolved when they refresh the page

app/models/user.rb
```diff
@@ -58,8 +58,8 @@ def as_json(_options = {})
...
+        maps: following.where(muted: false, followed_type: 'Map').to_a.map(&:followed_id)
```
devvmh: we could consider creating a scope for this, maybe called `active` or something, so the code would become

    map: following.active.where(followed_type: 'Map').to_a.map(&:followed_id)

app/services/follow_service.rb
devvmh: this should use follow.update, rather than update_attribute, so validations are run. then we need to check if it succeeded and handle validation errors

app/services/notification_service.rb
devvmh: probably just merge these two lines, especially if we make the `active` scope

app/models/user.rb
Connoropolous: right, makes sense

PR #1133, opened 2017-09-10
Topic Card Design changes (to include 'follows')

Connoropolous:
this exposes the metacode title into the main topic card view, which I think is useful

this fixes a bug where logged out users can't open the topic cards

it also adjusts the minimum height of the topic card, so that there's not a bunch of extra space where the desc should be.

Before:
screen shot 2017-09-10 at 5 24 43 pm

After:
screen shot 2017-09-10 at 5 20 13 pm
screen shot 2017-09-10 at 5 19 40 pm
screen shot 2017-09-10 at 5 19 51 pm


app/assets/stylesheets/application.scss.erb
```diff
@@ -3150,11 +3150,15 @@ script.data-gratipay-username {
...
-    text-align: center;
-    height: 48px;
-    line-height: 48px;
-    border-top: 1px solid #BDBDBD;
-    background: #FFF;
+    height: 24px;
+    line-height: 24px;
     cursor: pointer;
-    font-family: din-regular;
+    font-family: helvetica;
```
devvmh: do we need a fallback font?

app/assets/stylesheets/base.scss.erb
```diff
@@ -213,37 +205,52 @@
...
-      top: 1px;
-      width: 46px;
-      height: 46px;
-      background-size:46px 46px;
+      position: absolute;
+      left: -18px;
+      top: 6px;
+      width: 36px;
+      height: 36px;
+      background-size:36px 36px;
```
devvmh: some of this could turn into variables, but only if it makes sense. I'd like to see less and less magic numbers in css and more stuff based on proportions of other stuff.... but only if it's helpful. so if using a variable doesn't make sense here, don't do it, but I'm commenting so you can consider it

app/assets/stylesheets/base.scss.erb
```diff
@@ -213,37 +205,52 @@
...
+  .linkItem {
+    float: left;
+    z-index: 1;
+    position: relative;
+    color: rgba(66, 66, 66, 0.6);
+    font-size: 14px;
+    line-height: 14px;
+
+    a {
+      color: rgba(66, 66, 66, 0.6);
```
devvmh: color could be a named variable

app/assets/stylesheets/base.scss.erb
```diff
@@ -306,18 +295,18 @@
...
         content: '';
         position: absolute;
         top: 26px;
-        left: 10px;
+        right: 10px;
         margin-top: -30px;
         width: 0;
         height: 0;
-        border-bottom: 4px solid #000000;
+        border-bottom: 4px solid #8A8A8A;
```
devvmh: https://www.youtube.com/watch?v=IlqLLqOEKNo

app/assets/stylesheets/base.scss.erb
```diff
@@ -398,10 +389,10 @@
       content: '';
       position: absolute;
       margin-top: -8px;
-      margin-left: 6px;
+      right: 12px;
       width: 0;
       height: 0;
-      border-bottom: 4px solid black;
+      border-bottom: 4px solid #8A8A8A;
```
devvmh: but seriously i guess we could name this colour too since it's reused

app/assets/stylesheets/base.scss.erb
```diff
@@ -434,14 +429,10 @@
 }
 
 .yourTopic .mapPerm:hover, .yourEdge .mapPerm:hover {
-  background-image: url(<%= asset_data_uri('arrowperms_sprite.png') %>);
-  background-position: -32px 0;
```
devvmh: nice, does this mean we can remove this file from the repo?

app/assets/stylesheets/base.scss.erb
```diff
@@ -477,32 +468,34 @@ cursor: pointer;
...
+  top: 2px;
+  left: 4px;
   width: 16px;
   height: 16px;
   background-image: url(<%= asset_data_uri('arrowright_sprite.png') %>);
   background-repeat: no-repeat;
   background-position: 0 -32px;
+  display: inline-block;
+  -webkit-transform: rotate(90deg);
+  transform: rotate(90deg);
```
devvmh: remember my browser support PR? I know we're not there yet, but what set of browsers does these two transforms support? i'm assuming safari/chrome/opera are good, what about FF and Edge? I'm comfortable dropping IE11 if you want

This question is specific, but the answer depends on a more general question about our browser support strategy. And I guess our CSS support doesn't HAVE to match our JS, but it'd be nice to have a clear answer on how far to go to support different browsers for this sort of thing. 

I hope this comment makes sense. I can clarify if not

app/views/shared/_metacodeBgColors.html.erb
```diff
@@ -2,8 +2,8 @@
 <% Metacode.all.each do |m| %>
 	<% if m.color %>
 		<%= ".mbg" + m.id.to_s + "{" %>
-		<%= "background-color:" + m.color + " !important;" %>
+		<%= "color:" + m.color + " !important;" %>
```
devvmh: should we rename the file too?

frontend/src/components/MetacodeSelect.js
```diff
@@ -13,6 +13,7 @@ import PropTypes from 'prop-types'
 
 class MetacodeSelect extends Component {
   render = () => {
+    if (!this.props.metacodeSets) return null
```
devvmh: interesting, what case does this cover? the dev server? should we console.log some instructions about how to set it up in this case?

frontend/src/components/TopicCard/Desc.js
```diff
@@ -26,7 +26,7 @@ class MdTextArea extends RIETextArea {
 class Desc extends Component {
   render = () => {
     const descHTML = (!this.props.desc && this.props.authorizedToEdit)
-      ? '<p>Click to add description...</p>'
+      ? '<p class="emptyDesc">Edit the description...</p>'
```
devvmh: did you feel like some of this was easier because it's in react? I'm interested to hear if you think the developer experience has gotten better for you for this kind of sweeping change

frontend/src/components/TopicCard/Follow.js
```diff
@@ -3,16 +3,17 @@ import PropTypes from 'prop-types'
 
 class Follow extends Component {
   render = () => {
-    const { isFollowing, onTopicFollow } = this.props
-    return <div className='topicFollow' onClick={onTopicFollow}>
-      {isFollowing ? 'Unfollow' : 'Follow'}
+    const { ActiveMapper, isFollowing, onTopicFollow } = this.props
+    return <div className='topicFollow' onClick={() => ActiveMapper && onTopicFollow()}>
```
devvmh: I don't think it's really a bottleneck for us, so it probably doesn't matter, but we should consider a broader strategy for these types of functions. The best way to do this kind of thing is declare a function in the class using an arrow function so `this` is automatically bound (see https://medium.freecodecamp.org/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56 for a good discussion on the relative merits of different ways)

But this is concise and developer friendly so I don't think you need to change it here. But we should consider maybe someday refactoring our components to follow one pattern or the other, especially if we start thinking about JS performance at any time

devvmh: I think we'd want to set up some profiling to really dive into js perf

frontend/src/components/TopicCard/Follow.js
```diff
@@ -3,16 +3,17 @@ import PropTypes from 'prop-types'
...
+      {ActiveMapper ? isFollowing ? 'Unfollow' : 'Follow' : ''}
```
devvmh: this is too clever for me. I'd rather this be a function, honestly, with spelled out if statements. Even though the chained ternaries are pretty sexy lol

frontend/src/components/TopicCard/Info.js
```diff
@@ -0,0 +1,104 @@
+/* global $ */
+
+import React, { PropTypes, Component } from 'react'
```
devvmh: Here and any other new components you made, you'll get a console warning unless you remove `PropTypes` from this line and add this line:

     import PropTypes from 'prop-types'

Because I updated react

app/assets/stylesheets/base.scss.erb
Connoropolous: lol

app/assets/stylesheets/base.scss.erb
Connoropolous: looks like, from 'caniuse' that I don't even need the webkit one actually.

app/views/shared/_metacodeBgColors.html.erb
Connoropolous: yeah

frontend/src/components/MetacodeSelect.js
Connoropolous: no, its actually when the user is logged out. it fixes a current bug where logged out user can't open the topic card

frontend/src/components/TopicCard/Desc.js
Connoropolous: yeah 👍  👍  for sure

frontend/src/components/TopicCard/Follow.js
Connoropolous: I don't follow you here, *which* type of functions?

frontend/src/components/TopicCard/Follow.js
Connoropolous: yeah, will do. I rushed this line

frontend/src/components/TopicCard/Info.js
Connoropolous: kk, I will go through and update any outstanding

frontend/src/components/TopicCard/Follow.js
devvmh: Putting an arrow function in a prop; it has go reallocate memory or something for the function each time it renders

Connoropolous: oh i c i c

app/assets/stylesheets/base.scss.erb
Connoropolous: no, we still use it for the synapse card

Connoropolous: actually wait, no I'll remove it, cause I removed it from synapse card too

frontend/src/components/TopicCard/index.js
```diff
@@ -29,8 +29,8 @@ class ReactTopicCard extends Component {
     if (topic.authorizePermissionChange(currentUser)) classname += ' yourTopic'
 
     return (
-      <Draggable handle=".metacodeImage" defaultPosition={{x: 100, y: 100}}>
-        <div className="showcard mapElement mapElementHidden">
+      <Draggable handle='.metacodeImage' defaultPosition={{x: 100, y: 100}}>
+        <div className='showcard mapElement mapElementHidden'>
```
devvmh: is react ok with single quotes on props? i am if it is

app/assets/stylesheets/base.scss.erb
devvmh: awesome

frontend/src/components/TopicCard/index.js
Connoropolous: yep

app/assets/stylesheets/application.scss.erb
Connoropolous: yeah, added

PR #1137, opened 2017-09-17
make desc edit area larger and resizable

Connoropolous:
This is related to #1136
Except instead of it auto-resizing, based on some calculation, it just starts with a larger default, and allows resizing by the user. This is because calculating the number of rows it should is actually really hard. A combination of string length and line breaks

### DIFF
--- a/app/assets/stylesheets/base.scss.erb
+++ b/app/assets/stylesheets/base.scss.erb
@@ -139,14 +139,12 @@ $mid-gray-opacity: rgba(66, 66, 66, 0.6);
   font-family: helvetica, sans-serif;
   color: #424242;
   padding: 0;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/base.scss.erb
```diff
@@ -139,14 +139,12 @@ $mid-gray-opacity: rgba(66, 66, 66, 0.6);
   font-family: helvetica, sans-serif;
   color: #424242;
   padding: 0;
-  width: 100%;
+  width: 258px;
```
devvmh: is this OK on mobile?

Connoropolous: yeah

PR #1139, opened 2017-09-19
Right click/context menu redone in React

### LINKED ISSUE #1075: Convert entire right click menu to React component

Connoropolous:
fixes #1075 

todo 
- [x] pass in ReactApp.render as a callback function instead of importing and calling
- [x] improve render function of ContextMenu component by splitting it up into bunch of smaller pieces
- [x] uncomment the positionality code that adjusts for which part of the screen user right clicks in


frontend/src/Metamaps/GlobalUI/ReactApp.js
```diff
@@ -155,6 +156,29 @@ const ReactApp = {
       onTopicFollow: Topic.onTopicFollow
     }
   },
+  getContextMenuProps: function() {
+    const self = ReactApp
```
devvmh: can drop

frontend/src/Metamaps/Topic.js
```diff
@@ -156,7 +157,7 @@ const Topic = {
         })
       })
       if ($.isArray(nodes) && nodes.length > 1) {
```
devvmh: Are we using jquery much in this file? I wonder if we can switch to Array.isArray

frontend/src/Metamaps/Views/ContextMenu.js
```diff
@@ -0,0 +1,108 @@
...
+  pos: {x: 0, y: 0},
+  fetchingSiblingsData: false,
+  siblingsData: null,
+  selectNode: (node, pos) => {
+    ContextMenu.pos = pos
+    ContextMenu.clickedNode = node
+    ContextMenu.clickedEdge = null
+    ContextMenu.fetchingSiblingsData = false
+    ContextMenu.siblingsData = null
+    ReactApp.render()
```
devvmh: this makes me nervous. Something about accessing global objects in this way. The Control stuff doesn't so much, because it's hard to avoid.

Could we pass a render callback to each of these functions that calls ReactApp.render? I think that would be an improvement. At least it would make these more functional/testable

devvmh: I'm not sure if I should ask about including tests for some of these files or not ;)

frontend/src/Metamaps/Views/ContextMenu.js
```diff
@@ -0,0 +1,108 @@
...
+    Control.hideSelectedNodes()
+    ContextMenu.reset()
+  },
+  centerOn: (id) => {
+    Topic.centerOn(id)
+    ContextMenu.reset()
+  },
+  popoutTopic: (id) => {
+    ContextMenu.reset()
+    const win = window.open('/topics/' + id, '_blank')
```
devvmh: could use

    `/topics/${id}`

if you want

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,148 @@
...
+    super(props)
+    this.state = {
+      populateSiblingsSent: false
+    }
+  }
+
+  render () {
+    const { contextNode, contextPos, contextOnMetacodeSelect, metacodeSets,
+            contextSiblingsData, contextFetchSiblings,
+            contextPopulateSiblings, contextFetchingSiblingsData,
```
devvmh: in here, prefixing all these with "context" doesn't make much sense. I think they should be passed in like

     node={contextNode}

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,148 @@
...
+            <div className="rc-keyboard">Ctrl+M</div>
+        </li>
+        <li className="rc-delete ">
+          <div className="rc-icon" />Delete
+            <div className="rc-keyboard">Ctrl+D</div>
+        </li>
+        <li className="rc-popout">
+          <div className="rc-icon" />Open in new tab
+        </li>
+        <li className="rc-spacer"></li>
```
devvmh: I'd love to see these broken out into helper functions, just to make the render function smalller and easier to parse

devvmh: I do get this is a draft lol but when it's ready for prime time I mean

frontend/src/Metamaps/GlobalUI/ReactApp.js
```diff
@@ -155,6 +157,28 @@ const ReactApp = {
...
+    return {
+      // values
+      contextMenu: !!(ContextMenu.clickedNode || ContextMenu.clickedEdge),
+      contextNode: ContextMenu.clickedNode,
+      contextEdge: ContextMenu.clickedEdge,
+      contextPos: ContextMenu.pos,
+      contextFetchingSiblingsData: ContextMenu.fetchingSiblingsData,
+      contextSiblingsData: ContextMenu.siblingsData,
+      // functions
+      contextDelete: apply(ContextMenu.delete, render),
```
Connoropolous: we could/should potentially reuse this pattern elsewhere

Connoropolous:
alrighty @devvmh good for another round?

frontend/src/Metamaps/Views/ContextMenu.js
```diff
@@ -0,0 +1,108 @@
...
+      for (var key in data) {
+        data[key] = `${DataModel.Metacodes.get(key).get('name')} (${data[key]})`
+      }
+      ContextMenu.siblingsData = data
+      render()
+    }
+
+    $.ajax({
+      type: 'GET',
+      url: '/topics/' + id + '/relative_numbers.json?network=' + topicsString,
```
devvmh: this one would be great as a template string

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+    contextDelete: PropTypes.func,
+    contextRemove: PropTypes.func,
+    contextHide: PropTypes.func,
+    contextCenterOn: PropTypes.func,
+    contextPopoutTopic: PropTypes.func,
+    contextUpdatePermissions: PropTypes.func,
+    contextOnMetacodeSelect: PropTypes.func,
+    contextFetchSiblings: PropTypes.func,
+    contextPopulateSiblings: PropTypes.func
+  }
```
devvmh: you could throw in some isRequired if any are super duper required

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+
+  getPositionData = () => {
+    const { contextPos } = this.props
+    let extraClasses = []
+    const position = {}
+    const RIGHTCLICK_WIDTH = 300
+    const RIGHTCLICK_HEIGHT = 144 // this does vary somewhat, but we can use static
+    const SUBMENUS_WIDTH = 256
+    const MAX_SUBMENU_HEIGHT = 270
```
devvmh: these magic numbers are scary. I'm not sure if it's better to have them here or in css. At least in css you could in theory refactor to a proportion. 

I guess this is a refactor of existing functionality so I'll let it go for now

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+      extraClasses.push('moveMenusUp')
+    } else {
+      position.top = contextPos.y
+    }
+    return {
+      pos: {
+        top: position.top && position.top + 'px',
+        bottom: position.bottom && position.bottom + 'px',
+        left: position.left && position.left + 'px',
+        right: position.right && position.right + 'px'
```
devvmh: someday we're gonna have https://github.com/tc39/proposal-optional-chaining and it's gonna be great

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+      extraClasses
+    }
+  }
+
+  hide = () => {
+    const { contextHide } = this.props
+    return <li className='rc-hide' onClick={contextHide}>
+      <div className='rc-icon' />Hide until refresh
+        <div className='rc-keyboard'>Ctrl+H</div>
+    </li>
```
devvmh: missing closing div?

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+  remove = () => {
+    const { contextRemove, map, currentUser } = this.props
+    const canEditMap = map && map.authorizeToEdit(currentUser)
+    if (!canEditMap) {
+      return null
+    }
+    return <li className='rc-remove' onClick={contextRemove}>
+      <div className='rc-icon' />Remove from map
+        <div className='rc-keyboard'>Ctrl+M</div>
+    </li>
```
devvmh: closing div

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  delete = () => {
+    const { contextDelete, map, currentUser } = this.props
+    const canEditMap = map && map.authorizeToEdit(currentUser)
+    if (!canEditMap) {
+      return null
+    }
+    return <li className='rc-delete' onClick={contextDelete}>
+      <div className='rc-icon' />Delete
```
devvmh: closing div

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  center = () => {
+    const { contextCenterOn, contextNode, topicId } = this.props
+    if (!(contextNode && topicId)) {
+      return null
+    }
+    return <li className='rc-center'
+      onClick={() => contextCenterOn(contextNode.id)}>
+      <div className='rc-icon' />Center this topic
```
devvmh: div

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+      <div className='rc-icon' />Delete
+      <div className='rc-keyboard'>Ctrl+D</div>
+    </li>
+  }
+
+  center = () => {
+    const { contextCenterOn, contextNode, topicId } = this.props
+    if (!(contextNode && topicId)) {
+      return null
+    }
```
devvmh: btw these early returns in each function are 👌 👌 👌 👌 👌

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+    </li>
+  }
+
+  permission = () => {
+    const { currentUser, contextUpdatePermissions } = this.props
+    if (!currentUser) {
+      return null
+    }
+    return <li className='rc-permission'>
+      <div className='rc-icon' />Change permissions
```
devvmh: div

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  popout = () => {
+    const { contextPopoutTopic, contextNode } = this.props
+    if (!contextNode) {
+      return null
+    }
+    return <li className='rc-popout'
+      onClick={() => contextPopoutTopic(contextNode.id)}>
+      <div className='rc-icon' />Open in new tab
```
devvmh: div

frontend/src/components/common/ContextMenu.js
Connoropolous: these are closed inline :)

devvmh: ohhhhhhh, all of these are self-closed divs

devvmh: Can we stick the text on another line for each one?

frontend/src/components/common/ContextMenu.js
Connoropolous: 🙏 

(I def agree. We could throw in a TODO about making them dynamic, because the contextmenu is bound to shrink and grow)

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+    if (!currentUser) {
+      return null
+    }
+    return <li className='rc-metacode'>
+      <div className='rc-icon' />Change metacode
+      <div id='metacodeOptionsWrapper'>
+        <MetacodeSelect
+          onMetacodeSelect={id => {
+            contextOnMetacodeSelect(contextNode && contextNode.id, id)
+          }}
```
devvmh: FYI this is another example of the arrow function that gets stored in memory or something on every render. If we have problems with our right click menu's performance lol we can re-evaluate. No action needed on this comment for now

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+    if (!(contextNode && topicId)) {
+      return null
+    }
+    return <li className='rc-siblings'
+        onMouseOver={populateSiblings}>
+      <div className='rc-icon' />Reveal siblings
+      <ul id='fetchSiblingList'>
+        <li className='fetchAll'
+          onClick={() => contextFetchSiblings(contextNode)}>All
```
devvmh: yeah I like it better if "All" is on its own line even here

frontend/src/components/common/ContextMenu.js
Connoropolous: Yeahhhh

frontend/src/components/common/ContextMenu.js
devvmh: i guess i can't always get what i want though haha

frontend/src/components/common/ContextMenu.js
```diff
@@ -0,0 +1,234 @@
...
+        {this.center()}
+        {this.popout()}
+        {showSpacer && <li className='rc-spacer' />}
+        {this.permission()}
+        {this.metacode()}
+        {this.siblings()}
+      </ul>
+    </div>
+  }
+}
```
devvmh: beautiful

frontend/src/components/common/ContextMenu.js
devvmh: what did you think about not prefixing the names with "context" for all these props?

frontend/src/components/common/ContextMenu.js
Connoropolous: Sure

package.json
```diff
@@ -21,6 +21,7 @@
   "homepage": "https://github.com/metamaps/metamaps#readme",
   "dependencies": {
     "ajaxq": "0.0.7",
+    "async": "2.5.0",
```
devvmh: if this is only used for `apply`, I believe there's a built in we can use instead

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

or I think actually the way you're currently using apply is more like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

devvmh: wow, this is actually different than I thought. It would be nice if you can somehow make this work though to avoid the dependency.

Connoropolous: The way that this version of `apply` works is different than either of those
https://caolan.github.io/async/docs.html#apply
I used it a lot at bitHound that's how I knew about it. Really good for functional js

frontend/src/components/common/ContextMenu.js
Connoropolous: yeah

frontend/src/components/common/ContextMenu.js
Connoropolous: I agree that in the future we should. The reason I am not right is because there is a weird/bad pattern we are using where many/most components get called with ALLL the props generated by ReactApp.getProps(). So having them namespaced is important. I could make each one explicit at the point where the ContextMenu gets rendered, but that'd be ugly ugly.

package.json
Connoropolous: @devvmh what's the advantage in your opinion of avoiding the dependency?

devvmh: If its useful and good lets add it.

Someday ill set up a profiler and look at our bundle size, then we can make decisions about including dependencies with actual data

PR #1142, opened 2017-09-20
Notifications Dropdown

Connoropolous:
- [x] improve loading styling (add a loading component)
- [x] add hover states to buttons in the dropdown box
- [x] create a version for the empty case where user has no notifs

### DIFF
--- a/frontend/src/Metamaps/GlobalUI/Notifications.js
+++ b/frontend/src/Metamaps/GlobalUI/Notifications.js
@@ -15,14 +15,22 @@ const Notifications = {
       }
     })
   },
+  incrementUnread: (render) => {
+    Notifications.unreadNotificationsCount++
+    render()
+  },
+  decrementUnread: (render) => {
+    Notifications.unreadNotificationsCount--
+    render()
+  },
   markAsRead: (render, id) => {
     const n = Notifications.notifications.find(n => n.id === id)
     $.ajax({
-      url: `/notifications/${id}/mark_read`,
+      url: `/notifications/${id}/mark_read.json`,
       method: 'PUT',
       success: function(r) {
         if (n) {
-          Notifications.unreadNotificationsCount--
+
           n.is_read = true
           render()
         }
@@ -35,7 +43,7 @@ const Notifications = {
   markAsUnread: (render, id) => {
     const n = Notifications.notifications.find(n => n.id === id)
     $.ajax({
-      url: `/notifications/${id}/mark_unread`,
+      url: `/notifications/${id}/mark_unread.json`,
       method: 'PUT',
       success: function() {
         if (n) {
--- a/frontend/src/Metamaps/GlobalUI/Notifications.js
+++ b/frontend/src/Metamaps/GlobalUI/Notifications.js
@@ -1,11 +1,51 @@
+import GlobalUI from './index'
+
 const Notifications = {
   notifications: null,
+  unreadNotificationsCount: 0,
+  init: serverData => {
+    Notifications.unreadNotificationsCount = serverData.unreadNotificationsCount
+  },
   fetch: render => {
     $.ajax({
       url: '/notifications.json',
       success: function(data) {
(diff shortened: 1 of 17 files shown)


frontend/src/Metamaps/GlobalUI/Notifications.js
```diff
@@ -0,0 +1,54 @@
+import GlobalUI from './index'
+
+const Notifications = {
+  notifications: null,
+  unreadNotificationsCount: 0,
```
devvmh: it would be great if this was a function rather than a global variable like this. isn't that doable?

Connoropolous: You mean unreadNotificationsCount

Connoropolous: ?

devvmh: Yeah

Connoropolous: @devvmh a function how? and where does the value get stored? 
locally, and then just return it within the function?

devvmh:
```
unreadNotificationsCount: (notifications) => {
  return notifications.filter(n => !n.is_read).length
},
```

devvmh: Does that work? I just want to avoid storing global state

Connoropolous: no, that doesn't work because the notifications array can't be depended on for completeness. 
at any given time only a certain page of all the notifications may be stored locally. That's why we had the number occurring separately and coming down from the server

devvmh: Ok,  too bad.  Maybe I'll take a look sometime

frontend/src/components/Notification.js
```diff
@@ -26,10 +27,58 @@ class Notification extends Component {
...
+        result += notification.subject
+    }
+    return {__html: result}
+  }
+
   getDate = () => {
     const { notification: {created_at} } = this.props
     const months = ['Jan','Feb','Mar','Apr','May','Jun',
-                    'Jul','Aug','Sep','Oct','Nov','Dec']
+    'Jul','Aug','Sep','Oct','Nov','Dec']
```
devvmh: this lost indentation

frontend/src/components/Notification.js
```diff
@@ -47,28 +96,31 @@ class Notification extends Component {
...
         {notification.is_read && <div onClick={this.markAsUnread}>
           mark unread
         </div>}
       </div>
-      <div className="notification-date">
+      <div className='notification-date'>
         {this.getDate()}
       </div>
-      <div className="clearfloat"></div>
+      <div className='clearfloat'></div>
```
devvmh: I don't view this as blocking, but I think JSX props should use double quotes by default. It's a more common style, and it more closely mirrors HTML

frontend/src/components/App/NotificationBox.js
```diff
@@ -0,0 +1,55 @@
...
+        {empty && <li className='notificationsEmpty'>
+          You have no notifications. <br />
+          More time for dancing.
+        </li>}
+        {notifications && notifications.slice(0, 10).map(n => {
+          return <Notification notification={n}
+            markAsRead={markAsRead}
+            markAsUnread={markAsUnread}
+            key={`notification-${n.id}`} />
+        })}
```
devvmh: Could replace linea 36-45 with:

```
{empty ? (
  <li ...</li>
) : (
  notifications.slice(...).map...
)}
```

Connoropolous: 👍

Connoropolous: I have a third revision for you to check out

PR #1144, opened 2017-09-23
hide "learn markdown" unless you're editing the topic card desc

devvmh:


Connoropolous:
Did you test this? Are the divs nested correctly such that this works?

devvmh:
Yeah i tested this in one topic in localhost.

I suppose it'll break if we refactor the react component at all

PR #1150, opened 2017-10-14
run perms:fix on assets:precompile

devvmh:


lib/tasks/extensions.rake
```diff
@@ -3,8 +3,13 @@ namespace :assets do
...
+  end
+
+  task :production_ready do
     system 'bin/build-apidocs.sh' if Rails.env.production?
+    Rake::Task['perms:fix'].invoke if Rails.env.production?
   end
 end
 
 Rake::Task[:'assets:precompile'].enhance([:'assets:js_compile'])
+Rake::Task[:'assets:precompile'].enhance([:'assets:production_ready'])
```
Connoropolous: what's this line? how was it working before if it wasn't there?

devvmh: before i did three `system` commands in a task called js_compile. You can run `rake assets:js_compile` to run those three commands on the current develop.

In this PR I split it into two tasks. So after merging you could run `rake assets:production_ready` to do perms:fix and bin/build-apidocs.sh

Then I "enhance" the task with both other tasks instead of the one i had before

Connoropolous: cool

PR #1152, opened 2017-10-15
perms:fix needs to run after assets:precompile

Connoropolous:
I tested this on prod and it worked

### DIFF
--- a/lib/tasks/extensions.rake
+++ b/lib/tasks/extensions.rake
@@ -11,5 +11,9 @@ namespace :assets do
   end
 end
 
+# run before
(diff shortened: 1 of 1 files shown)


lib/tasks/extensions.rake
```diff
@@ -11,5 +11,9 @@ namespace :assets do
   end
 end
 
+# run before
 Rake::Task[:'assets:precompile'].enhance([:'assets:js_compile'])
-Rake::Task[:'assets:precompile'].enhance([:'assets:production_ready'])
+# run after
+Rake::Task[:'assets:precompile'].enhance do
+  Rake::Task[:'assets:production_ready'].invoke
+end
```
devvmh: perfect

PR #1155, opened 2017-10-17
fix map activity emails

Connoropolous:
the reason we weren't seeing the error messages is being they would be showing up in the DelayedJob logs, since delayedjob is the one actually sending them 
(because we use deliver_later in the rake task)

and the reason why I didn't catch the template error is because in my test case, all the different possible changes occurred, not just some of them. If any type of change had not occurred, I would have realized that this would fail in that situation


Connoropolous:
I should write tests, wah wah

Connoropolous:
@devvmh apparently the delayedjob logs were... nowhere? no they'll be somewhere

Issue #1146, opened 2017-10-04
Map Info Box bug: closes itself when you click within the box


Connoropolous:
this is likely due to some erroneous click handler

devvmh:
I can't replicate this on Opera/macOS

Connoropolous:
are you logged in? I get it as a logged in user

devvmh:
Huh. Logged in on your public map (https://metamaps.cc/maps/3179) I don't get it, but logged in on my map (https://metamaps.cc/maps/1349) I can replicate

devvmh:
just changed https://metamaps.cc/maps/1349 to public so you can try if you want

Connoropolous:
Ok, I think I've determined it. It doesn't do it when you first load a map, but does it once you navigate HOME, and then BACK to a map.

PR #1160, opened 2017-10-26
MapInfoBox was closing annoyingly after you left initial map then opened new one

Connoropolous:
this is the same as on the initial page load (what happens in the 'init' function)

### DIFF
--- a/frontend/src/Metamaps/Map/InfoBox.js
+++ b/frontend/src/Metamaps/Map/InfoBox.js
@@ -116,7 +116,9 @@ const InfoBox = {
   },
   attachEventListeners: function() {
     var self = InfoBox
-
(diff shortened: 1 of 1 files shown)


devvmh (commented):
Why isn't this being removed from somewhere else?

Connoropolous:
cause it's the intended behaviour. We have a click event listener on 'body' which is designed to close the mapInfoBox, but mapInfoBox needs to stopPropogation to keep from closing itself when you click around on it. 

Similar to what we do in react with clickOutsideDiv

Connoropolous:
mapInfoBox unfortunately is not in react yet. It's up next

devvmh:
Ok lgtm ship it

Connoropolous:
fixes #1146

PR #1161, opened 2017-11-04
add normal/API endpoints for attachments

devvmh:
Split out some of the backend work from https://github.com/metamaps/metamaps/pull/1112

This PR just adds the controller endpoints and the API endpoints so the other
PR can focus on frontend work

TODO

- [ ] sanity test with `rails server`

### DIFF
--- a/app/policies/attachment_policy.rb
+++ b/app/policies/attachment_policy.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+class AttachmentPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      scope.where(attachable: TopicPolicy::Scope.new(user, Topic).resolve)
+    end
+  end
+
+  def index?
+    true
+  end
+
+  def create?
+    Pundit.policy(user, record.attachable).update?
+  end
+
+  def destroy?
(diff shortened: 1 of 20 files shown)


app/policies/attachment_policy.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class AttachmentPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      scope.where(attachable: TopicPolicy::Scope.new(user, Topic).resolve)
+    end
+  end
+
+  def index?
+    true
```
devvmh, line 11: this should be OK because of the scope above

app/policies/attachment_policy.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class AttachmentPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      scope.where(attachable: TopicPolicy::Scope.new(user, Topic).resolve)
```
devvmh, line 6: not sure how this will work once we are attaching to Message as well as Topic

Connoropolous (approved):
Looks good to me @devvmh

PR #1162, opened 2017-11-06
automatic rubocop fixes

devvmh:
just ran `rubocop -a`


devvmh:
OK, I found out that code climate lags behind the bleeding edge rubocop. So let's freeze rubocop on a version that matches code climate https://github.com/tootsuite/mastodon/issues/1758

I made that change and updated. I also reverted the line length change.

I'll do another PR that fixes line length issues following on this one

PR #1166, opened 2017-12-03
TopicCard/Permission.spec.js

devvmh:


frontend/src/components/TopicCard/Permission.js
```diff
@@ -15,10 +15,6 @@ class Permission extends Component {
     this.setState({selectingPermission: !this.state.selectingPermission})
   }
 
-  openPermissionSelect = () => {
```
Connoropolous, line 18: I guess this was unused, good catch

frontend/test/components/TopicCard/Permission.spec.js
```diff
@@ -0,0 +1,95 @@
...
+        permission: 'public', defer_to_map_id: null
+      })).to.equal(true)
+    })
+  })
+})
+
+describe('PermissionWrapper', function() {
+  const wrapper = shallow(<PermissionWrapper />)
+  it('wraps Permission component in onclickoutside', function() {
+    expect(wrapper.type()).to.equal(Permission)
```
Connoropolous, line 93: what is `wrapper.type()` ?

Connoropolous (approved):

frontend/test/components/TopicCard/Permission.spec.js
devvmh: I spend the whole time writing this pull request looking at the APIs for three libraries. Sinon, chai expectations, and enzyme shallow rendering. 

http://sinonjs.org/releases/v4.1.2/spies/

http://chaijs.com/api/bdd/

https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md

The wrappers are shallow rendered enzyme components which define a bunch of methods to help with assertions. 

type() returns a the component type of the root node, which will match the same type imported from somewhere else.  It's the way to say "when i render a Permission wrapper, the child should be a Permission component"

PR #1167, opened 2017-12-05
TopicCard/Desc.spec.js

devvmh:


devvmh:
code climate issue is an intentionally added TODO

PR #1172, opened 2018-02-06
factory_girl => factory_bot

devvmh:


Connoropolous:
@devvmh Is this because the name of the gem changed

devvmh:
Yeah i was looking for the docs and saw it changed. It seems like a positive change they put thought into and they're trying to get everyone to update so, why not!

https://robots.thoughtbot.com/factory_bot

PR #1175, opened 2018-03-17
finish Util.spec.js

devvmh:
title is a lie; I still need to finish the last function. Posting for early review


devvmh:
ok i have to fix the last test and then I found out there are four new functions so I'll try to get full coverage on this file before merging this PR

I won't write a test for isTester though

